PR #5, opened 2019-07-11
Bench adds

pascutto:


bench/dune
```diff
@@ -0,0 +1,9 @@
+(executable
+ (name main)
+ (libraries fmt index.unix))
+
+(alias
+ (name bench)
+ (deps main.exe)
+ (action
+  (run %{exe:main.exe})))
```
emillon: you can just say `(run ./main.exe)` and the dependency will be inferred automatically.

pascutto: Thanks !

PR #6, opened 2019-07-12
Detect log changes

pascutto:
Detect log changes (eg another process writing things) when the index is in read-only mode.


samoht:
Can you rebase, there are some conflicts.

PR #11, opened 2019-07-16
Have only one index file

pascutto:
For now this costs ~20% more time on insertions (merges included), and has shown no improvements for finds, thus marking as draft.


pascutto:
New implementation : performs as good as the former one for insertions, and gains ~10% on reads (30M sample). It's likely to be even better for bigger indexes (benchs in progess).
The main advantage is that it provides "free" fanout.

PR #12, opened 2019-07-17
Fix interpolation search for hash collisions

pascutto:


src/index.ml
```diff
@@ -199,8 +199,20 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+      else
+        let h_e = K.hash e.key in
+        if h_e <> h_key then None else search op off
+    in
+    match search Int64.add off with None -> search Int64.sub off | o -> o
+
   let interpolation_search t i key =
-    let hashed_key = float_of_int (K.hash key) in
+    let hashed_key = K.hash key in
+    let hashed_keyf = float_of_int (K.hash key) in
```
Julow: You can use `hashed_key` instead of recomputing the hash.

src/index.ml
```diff
@@ -214,19 +226,22 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
                 ( (high -. low)
-                *. (hashed_key -. lowest_hash)
+                *. (hashed_keyf -. lowest_hash)
                 /. (highest_hash -. lowest_hash) )
             in
             let off = low +. doff -. mod_float doff entry_sizef in
             let offL = Int64.of_float off in
             let e = get_entry t i offL in
+            let hashed_e = K.hash e.key in
             if K.equal e.key key then Some e.value
```
Julow: You could perform `K.equal` only when the hashes are equal. `hashed_e` must be computed for every iterations except the last one, anyway.

src/index.ml
```diff
@@ -214,19 +226,22 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
           let highest_entry = get_entry_iff_needed t i high highest_entry in
           let lowest_hash = float_of_int (K.hash lowest_entry.key) in
           let highest_hash = float_of_int (K.hash highest_entry.key) in
-          if lowest_hash > hashed_key || highest_hash < hashed_key then None
+          if lowest_hash > hashed_keyf || highest_hash < hashed_keyf then None
```
Julow: This comparison should work with ints. Float values are only used in computing `doff`, so you could move the conversions there.

pascutto:
Thanks for reviewing !

PR #20, opened 2019-07-25
Refactoring : Have a separate module for the fanout

pascutto:
This will ease further improvements.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -80,6 +80,40 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
 
   let entry_sizeL = Int64.of_int entry_size
 
+  module Fan = struct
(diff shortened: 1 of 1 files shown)


src/index.ml
```diff
@@ -80,6 +80,40 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
   let entry_sizeL = Int64.of_int entry_size
 
+  module Fan = struct
+    type t = { size : int; fans : int64 array; mask : int; shift : int }
+
+    let v n =
+      let size = n in
+      let nb_fans = 1 lsl size in
+      let fans = Array.make nb_fans (-1L) in
+      let shift = K.hash_size - size in
```
Julow: You could take `hash_size` as an argument so you don't depend on anything. `shift` is stored in `t` so this is free.

PR #21, opened 2019-07-25
Fan export

pascutto:
This is built on top of #20.
It writes the fan out when merging so that read only and restarting instances can read this (small) file instead of the whole index.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -80,6 +80,40 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
 
   let entry_sizeL = Int64.of_int entry_size
 
+  module Fan = struct
+    type t = { size : int; fans : int64 array; mask : int; shift : int }
+
+    let v n =
+      let size = n in
+      let nb_fans = 1 lsl size in
+      let fans = Array.make nb_fans (-1L) in
+      let shift = K.hash_size - size in
+      let mask = (nb_fans - 1) lsl shift in
+      { size; fans; mask; shift }
+
+    let fan t h = (h land t.mask) lsr t.shift
+
+    let clear t = Array.fill t.fans 0 (Array.length t.fans) (-1L)
+
+    let search t h =
+      let fan = fan t h in
+      let low = if fan = 0 then 0L else t.fans.(fan - 1) in
+      (low, t.fans.(fan))
+
+    let update t hash off =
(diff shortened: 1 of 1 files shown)


src/index.ml
```diff
@@ -112,6 +112,57 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+
+    external get_64 : string -> int -> int64 = "%caml_string_get64"
+
+    external swap64 : int64 -> int64 = "%bswap_int64"
+
+    let encode_int64 i =
+      let set_uint64 s off v =
+        if not Sys.big_endian then set_64 s off (swap64 v) else set_64 s off v
+      in
+      let b = Bytes.create 8 in
```
samoht: you can allocate that buffer only once outside of the function. As you do a copy (with `Bytes.to_string`) that should be safe (unless that code is used in threads...)

src/index.ml
```diff
@@ -426,7 +468,13 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
       |> EntrySet.elements
     in
     merge_with log t tmp;
+    let tmp_fan_path = t.root // "tmp" // "fan" in
+    let fan_path = t.root // "fan" in
+    let tmp_fan = IO.v ~fresh:true ~readonly:false ~generation tmp_fan_path in
+    Fan.export t.fan_out tmp_fan;
+    let fan = IO.v ~fresh:true ~readonly:false ~generation fan_path in
     IO.rename ~src:tmp ~dst:t.index;
+    IO.rename ~src:tmp_fan ~dst:fan;
```
samoht: beware: this is not atomic anymore. What happens it the program crashes between line 476 and 477?

samoht:
If possible, I think it would be better if we keep all of these information into the same file, e.g. put the fanout at the beginning of the file.

pascutto:
I don't think we can do this with the current append-only IO, can we ?

src/index.ml
samoht: (actually this is a very bad idea in the context of multi-threaded programs :p)

samoht:
Can't you serialise the fan-out first and then add the entries? Or keep some space to fill it at the end?

pascutto:
No satisfying solution for now. Closing.

PR #24, opened 2019-08-01
Don't allow negative offsets in the fanout

pascutto:
Now the fanout contains -1 everywhere when nothing was added to it and non negative numbers otherwise.
The issue was when some offsets are added to the fanout but some fans (at the begining of the fanout) remained empty, then we could get negative (-1) offsets when querying it.


samoht:
Can you add a test for this (tricky) behavior?

samoht:
Merging as it fixes the tests I was running, but please add tests :-)

PR #25, opened 2019-08-02
Do not allow iter_io to go further than IO.offset

pascutto:
`iter_io f io` could go further than `IO.offset io` when the page size was larger than the actually remaining valid offset. This causes problems when data actually exists beyond `IO.offset io`.
Should fix issues in https://github.com/mirage/irmin/pull/826

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -153,12 +153,16 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
 
   let index_path root = root // "index" // "index"
 
+  let page_size = Int64.mul entry_sizeL 1_000L
+
   let iter_io_off ?(min = 0L) ?max f io =
     let max = match max with None -> IO.offset io | Some m -> m in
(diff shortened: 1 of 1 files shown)


src/index.ml
```diff
@@ -153,12 +153,16 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+
   let iter_io_off ?(min = 0L) ?max f io =
     let max = match max with None -> IO.offset io | Some m -> m in
     let rec aux offset =
-      if offset >= max then ()
+      let remaining = Int64.sub max offset in
+      if remaining <= 0L then ()
       else
-        let raw = Bytes.create (entry_size * 1_000) in
+        let size = Pervasives.min remaining page_size in
```
samoht: using `Pervasives` directly will break on 4.08

pascutto: Thanks !

PR #29, opened 2019-08-06
adding alcotests

icristescu:
Adding alcotests including tests with lwt threads, and testing the combinations of shared, readonly, fresh.

### DIFF
--- a/test/unix/main.ml
+++ b/test/unix/main.ml
@@ -52,9 +52,7 @@ let page_size = 2
 
 let pool_size = 2
 
-let fan_out_size = 1
-
-let t = Index.v ~fresh:true ~log_size ~fan_out_size index_name
+let t = Index.v ~fresh:true ~log_size index_name
 
 let l = ref (List.init index_size (fun _ -> (Key.v (), Value.v ())))
 
@@ -112,23 +110,18 @@ let find_present_live () = test_find_present t
 let find_absent_live () = test_find_absent t
 
 let find_present_restart () =
-  test_find_present (Index.v ~fresh:false ~log_size ~fan_out_size index_name)
+  test_find_present (Index.v ~fresh:false ~log_size index_name)
 
 let find_absent_restart () =
-  test_find_absent (Index.v ~fresh:false ~log_size ~fan_out_size index_name)
+  test_find_absent (Index.v ~fresh:false ~log_size index_name)
 
 let replace_live () = test_add t
 
-let replace_restart () =
-  test_add (Index.v ~fresh:false ~log_size ~fan_out_size index_name)
+let replace_restart () = test_add (Index.v ~fresh:false ~log_size index_name)
 
 let readonly () =
-  let w =
-    Index.v ~fresh:true ~readonly:false ~log_size ~fan_out_size index_name
-  in
-  let r =
-    Index.v ~fresh:false ~readonly:true ~log_size ~fan_out_size index_name
-  in
+  let w = Index.v ~fresh:true ~readonly:false ~log_size index_name in
+  let r = Index.v ~fresh:false ~readonly:true ~log_size index_name in
   List.iter (fun (k, v) -> Index.add w k v) !l;
   Index.flush w;
   List.iter
--- a/bench/main.ml
+++ b/bench/main.ml
@@ -50,9 +50,7 @@ let index_name = "hello"
 
 let log_size = 500_000
 
-let fan_out_size = 13
-
-let t = Index.v ~fresh:true ~log_size ~fan_out_size index_name
+let t = Index.v ~fresh:true ~log_size index_name
 
 let pp_stats ppf (count, max) =
   Fmt.pf ppf "\t%4dk/%dk" (count / 1000) (max / 1000)
--- a/src/fan.ml
+++ b/src/fan.ml
@@ -1,0 +1,46 @@
+type t = {
+  hash_size : int;
+  entry_size : int;
+  size : int;
+  fans : int64 array;
+  mask : int;
+  shift : int;
+}
+
+let log2 a = log a /. log 2.
+
+let v ~hash_size ~entry_size n =
+  let entry_sizef = float_of_int entry_size in
+  let entries_per_page = 4096. /. entry_sizef in
+  let entries_fan = float_of_int n /. (entries_per_page *. entries_per_page) in
+  let size = max 0 (int_of_float (ceil (log2 entries_fan))) in
+  let nb_fans = 1 lsl size in
+  let shift = hash_size - size in
+  {
+    hash_size;
+    entry_size;
+    size;
+    fans = Array.make nb_fans 0L;
+    mask = (nb_fans - 1) lsl shift;
+    shift;
+  }
+
+let fan t h = (h land t.mask) lsr t.shift
+
+let search t h =
+  let fan = fan t h in
+  let low = if fan = 0 then 0L else t.fans.(fan - 1) in
+  (low, t.fans.(fan))
+
+let update t hash off =
+  let fan = fan t hash in
+  t.fans.(fan) <- off
+
+let finalize t =
+  let rec loop curr i =
+    if i = Array.length t.fans then ()
+    else (
+      if t.fans.(i) = 0L then t.fans.(i) <- curr;
+      loop t.fans.(i) (i + 1) )
+  in
+  loop 0L 0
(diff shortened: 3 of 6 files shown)


test/unix/main.ml
```diff
@@ -94,6 +118,24 @@ let test_find_absent t =
   in
   loop index_size
 
+let test_no_extra_values_added t =
+  Index.iter
+    (fun k v ->
+      try
+        let l' = List.find_all (fun (k', _) -> String.equal k' k) !l in
+        if not (List.mem (k, v) l') then
```
pascutto: These two list calls can be replaced with a single `List.exists`, right ?

test/unix/main.ml
```diff
@@ -107,16 +149,68 @@ let test_add t =
...
+    Alcotest.fail (Printf.sprintf "Index readonly cannot be fresh.")
+  with _ -> ()
+
+let fail_restart_shared () =
+  try
+    let _ =
+      Index.v ~shared:true ~readonly:true ~log_size ~fan_out_size index_name
+    in
+    Alcotest.fail (Printf.sprintf "Index readonly cannot be shared.")
+  with _ -> ()
```
pascutto: I'm not sure what this is testing, I think you might want to use `Alcotest.check_raises`.

test/unix/main.ml
```diff
@@ -107,16 +149,68 @@ let test_add t =
...
+
+let fail_restart () =
+  try
+    let _ =
+      Index.v ~fresh:false ~readonly:true ~log_size ~fan_out_size "new_index"
+    in
+    Alcotest.fail (Printf.sprintf "Index readonly for non existing file.")
+  with _ -> ()
```
pascutto: Same as above.

test/unix/main.ml
```diff
@@ -69,6 +69,30 @@ let rec random_new_key () =
...
+        Fmt.(styled `Magenta string)
+        (Logs.Src.name src) Logs_fmt.pp_header (level, h)
+    in
+    msgf @@ fun ?header ?tags fmt ->
+    with_stamp header tags k fmt
+  in
+  { Logs.report }
+
+let () =
+  Logs.set_level (Some Logs.Debug);
```
CraigFe: This initialisation should probably be done in the same top-level function that runs the tests?

test/unix/main.ml
```diff
@@ -143,23 +238,86 @@ let readonly () =
...
+    Index.v ~shared:false ~fresh:false ~readonly:true ~log_size ~fan_out_size
+      index_name
+  in
+  let wp = Lwt_list.iter_p (fun (k, v) -> Lwt.return (Index.add w k v)) !l in
+  let p =
+    Lwt.bind
+      (Lwt.bind wp (fun _ -> Lwt.return (Index.flush w)))
+      (fun _ -> test_find_present_lwt r)
+  in
+  Lwt_main.run p
```
CraigFe: If you use `Alcotest_lwt.test_case` instead, you can avoid a call to `Lwt_main.run` here (since that `test_case` expects a `unit Lwt.t`)

test/unix/main.ml
```diff
@@ -143,23 +238,86 @@ let readonly () =
...
+  let w =
+    Index.v ~fresh:true ~readonly:false ~log_size ~fan_out_size index_name
+  in
+  let r =
+    Index.v ~shared:false ~fresh:false ~readonly:true ~log_size ~fan_out_size
+      index_name
+  in
+  let wp = Lwt_list.iter_p (fun (k, v) -> Lwt.return (Index.add w k v)) !l in
+  let p =
+    Lwt.bind
```
CraigFe: I suggest using the infix synonym for bind (`>>=`), to let this read more sequentially. It's common to open `Lwt.Infix` and use those operators instead of the standard functions.

test/unix/main.ml
```diff
@@ -143,23 +238,86 @@ let readonly () =
...
+      (Lwt.bind wp (fun _ -> Lwt.return (Index.flush w)))
```
CraigFe: This can be a `map` (or even better an infix `>|=` operator) to avoid needing `Lwt.return`.

test/unix/main.ml
```diff
@@ -107,16 +149,68 @@ let test_add t =
...
+  test_mem (Index.v ~fresh:false ~log_size ~fan_out_size index_name)
+
+let fail_restart_not_shared () =
+  try
+    let _ =
+      Index.v ~shared:false ~fresh:true ~readonly:true ~log_size ~fan_out_size
+        index_name
+    in
+    Alcotest.fail (Printf.sprintf "Index readonly cannot be fresh.")
+  with _ -> ()
```
CraigFe: I think you probably want to be using `Alcotest.check_raises` for these tests that expect an exception. It has the advantage that you can make sure you're getting the right exception.

test/unix/main.ml
```diff
@@ -107,16 +149,68 @@ let test_add t =
...
+let membership_restart () =
+  test_mem (Index.v ~fresh:false ~log_size ~fan_out_size index_name)
+
+let fail_restart_not_shared () =
+  try
+    let _ =
+      Index.v ~shared:false ~fresh:true ~readonly:true ~log_size ~fan_out_size
+        index_name
+    in
+    Alcotest.fail (Printf.sprintf "Index readonly cannot be fresh.")
```
```suggestion
    Alcotest.fail "Index readonly cannot be fresh."
```
This `sprintf` isn't doing anything.

test/unix/main.ml
```diff
@@ -143,23 +238,86 @@ let readonly () =
               (Printf.sprintf "Wrong insertion: %s value is missing." v))
     !l
 
+let readonly_no_flush () =
```
pascutto: Actually I think this is not part of the requirements that this should pass.
With the current implementation it's not passing for every situation (try playing with `index_size` for instance.
Sticking to the current requirement, ie reading after a flush must be correct is enough for now.

test/unix/main.ml
```diff
@@ -143,23 +238,86 @@ let readonly () =
...
+    Index.v ~shared:false ~fresh:false ~readonly:true ~log_size ~fan_out_size
+      index_name
+  in
+  try
+    let () = Index.add r (Key.v ()) (Value.v ()) in
+    Alcotest.fail (Printf.sprintf "Index readonly cannot write.")
+  with _ -> ()
+
+let test_clear () =
+  let () = Index.clear t in
```
CraigFe: I'm uncomfortable with this test depending on a global `t`. It suggests that the tests aren't robust to being selectively executed, but perhaps you don't want to initialise a new index for some reason?

test/unix/main.ml
```diff
@@ -143,23 +238,86 @@ let readonly () =
...
+  Index.iter
+    (fun _ _ -> Alcotest.fail (Printf.sprintf "Indexed not cleared"))
+    t
+
```
pascutto: You may also test that trying to find anything that was previously added (ie in `!l`) fails.

test/unix/main.ml
```diff
@@ -143,23 +238,86 @@ let readonly () =
...
+  in
+  try
+    let () = Index.add r (Key.v ()) (Value.v ()) in
+    Alcotest.fail (Printf.sprintf "Index readonly cannot write.")
+  with _ -> ()
+
+let test_clear () =
+  let () = Index.clear t in
+  Index.iter
+    (fun _ _ -> Alcotest.fail (Printf.sprintf "Indexed not cleared"))
```
```suggestion
    (fun _ _ -> Alcotest.fail "Indexed not cleared")
```

test/unix/main.ml
```diff
@@ -107,16 +149,68 @@ let test_add t =
...
+
 let find_present_live () = test_find_present t
 
 let find_absent_live () = test_find_absent t
 
+let no_extra_values_added () = test_no_extra_values_added t
+
+let membership () = test_mem t
+
+let find_present_live_lwt () = Lwt_main.run (test_find_present_lwt t)
```
```suggestion
let find_present_live_lwt _ () = test_find_present_lwt t
```

Use `Alcotest_lwt.test_case` for tests inside the Lwt monad to avoid needing to call `Lwt_main.run`. You'd also need to avoid the Lwt_switch.t that it offers you.

test/unix/main.ml
pascutto: Actually I believe the tests depend on each other, so they can't be run separately, witch was a pretty bad design from me. It probably needs a separate PR to handle this.

CraigFe: Not a big deal, but my experiences from debugging Irmin tests are that it's nice to leverage individual test execution where possible. Perhaps this is a situation that would benefit from an explicit 'test with context' feature in Alcotest.

PR #32, opened 2019-08-08
Drop `~fan_out_size` and automatically compute it instead

pascutto:
This also passes #29 tests.

### DIFF
--- a/src/fan.mli
+++ b/src/fan.mli
@@ -8,4 +8,4 @@ val search : t -> int -> int64 * int64
 
 val update : t -> int -> int64 -> unit
 
-val flatten : t -> unit
+val finalize : t -> unit
--- a/src/fan.mli
+++ b/src/fan.mli
@@ -2,8 +2,6 @@ type t
 
 val v : hash_size:int -> entry_size:int -> int -> t
 
-val clear : t -> unit
-
 val search : t -> int -> int64 * int64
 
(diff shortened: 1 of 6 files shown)


src/fan.mli
```diff
@@ -0,0 +1,17 @@
...
+val search : t -> int -> int64 * int64
+(** [search t hash] is the interval of offsets in witch [hash] is, if
+    present. *)
+
+val update : t -> int -> int64 -> unit
+(** [update t hash off] updates [t] so that [hash] is registered to be at
+    offset [off]. *)
+
+val finalize : t -> unit
+(** Finalizes the update of the fanout. This is mendatory before any [search]
```
CraigFe:
```suggestion
(** Finalizes the update of the fanout. This is mandatory before any [search]
```

src/fan.mli
```diff
@@ -0,0 +1,17 @@
+type t
+
+val v : hash_size:int -> entry_size:int -> int -> t
+(** [v ~hash_size ~entry_size n] creates a fan_out for an index with [hash_size]
+    and [entry_size], containing [n] elements. *)
+
+val search : t -> int -> int64 * int64
+(** [search t hash] is the interval of offsets in witch [hash] is, if
```
CraigFe:
```suggestion
(** [search t hash] is the interval of offsets containing [hash], if
```

PR #33, opened 2019-08-08
Tests

icristescu:
Recovering the old PR #29

### DIFF
--- a/test/unix/main.ml
+++ b/test/unix/main.ml
@@ -1,3 +1,4 @@
+module I = Index
 open Common
 
 let ( // ) = Filename.concat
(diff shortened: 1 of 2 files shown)


test/unix/main.ml
```diff
@@ -92,6 +115,21 @@ let test_find_absent t =
   in
   loop index_size
 
+let test_no_extra_values_added t =
+  Index.iter
+    (fun k v ->
+      if not (List.exists (fun (k', v') -> k = k' && v = v') !l) then
+        Alcotest.fail
+          (Printf.sprintf "Wrong insertion of key, value: %s, %s." k v))
```
CraigFe: Alcotest also provides `Alcotest.failf` for taking formatted strings directly

test/unix/main.ml
```diff
@@ -92,6 +115,21 @@ let test_find_absent t =
...
+    t
+
+let test_mem t =
+  List.iter
+    (fun (k, _) ->
+      if not (Index.mem t k) then
+        Alcotest.fail (Printf.sprintf "Wrong insertion: %s key is missing." k))
```
```suggestion
        Alcotest.failf "Wrong insertion: %s key is missing." k)
```

test/unix/main.ml
```diff
@@ -103,25 +141,73 @@ let test_add t =
...
+  Alcotest.check_raises "Index readonly cannot be fresh." exn (fun () ->
+      let _ =
+        Index.v ~shared:false ~fresh:true ~readonly:true ~log_size index_name
+      in
+      ())
+
+let fail_restart_shared () =
+  let exn = RO_Not_Allowed in
+  Alcotest.check_raises "Index readonly cannot be shared." exn (fun () ->
+      let _ = Index.v ~shared:true ~readonly:true ~log_size index_name in
```
CraigFe: There is also `ignore: 'a -> unit` for this pattern, but it's a matter of personal taste.
```suggestion
     ignore @@ Index.v ~shared:true ~readonly:true ~log_size index_name
```

test/unix/main.ml
```diff
@@ -103,25 +141,73 @@ let test_add t =
...
+
 let replace_live () = test_add t
 
 let replace_restart () = test_add (Index.v ~fresh:false ~log_size index_name)
 
 let readonly () =
   let w = Index.v ~fresh:true ~readonly:false ~log_size index_name in
-  let r = Index.v ~fresh:false ~readonly:true ~log_size index_name in
+  let r =
+    Index.v ~shared:false ~fresh:false ~readonly:true ~log_size index_name
```
CraigFe: >  `~shared:false`

This looks a bit unintuitive to me; if you don't have shared here I assume it sets the `readonly` flag in the `w` instance? cc @pascutto

test/unix/main.ml
pascutto: I'd prefer not to have `@@` if we can use parenthesis instead :)
Otherwise I agree.

test/unix/main.ml
icristescu: the `shared` flag is set to true by default, and thus in the `w` instance

pascutto: With the current implementation, if `shared` is set to `true`, then the instance is fetched in a hashtbl (searched by name), which means that `shared` + `readonly` would be the first instance created with `shared:true`, may it be readonly or not. (`shared` prevails).
I guess we would like the `readonly` to prevail, otherwise we could have permission errors when the process only have read access.
The solution brought in this PR is to make `shared` and `readonly` incompatible (which fits our use case so far). However it would probably be better to make these independant (in another PR).

samoht:
Regarding 322f13e, I'm wondering if it would be better to consider that opening a non-existing file in read-only will not have any side-effect on the disk but will appear like an empty file for the user. That would probably be less confusing...

samoht:
Also that would make it work if the underlying filesystem is really read-only!

pascutto:
I agree that actually writing a file when in readonly mode sounds weird.
However, I don't think this is a trivial change with the current implementation, maybe we should remove the test that fails because of this behaviour, merge the rest of them, and deal with this read-only opening in a different PR ? This PR has been opened for a while now, it would be nice to have it merged.

PR #35, opened 2019-08-09
Test size

icristescu:
Add an exception which is raised when trying to add keys and values of unexpected sizes.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -45,6 +45,10 @@ module type S = sig
 
   val mem : t -> key -> bool
 
+  exception Invalid_Key_Size of key
+
+  exception Invalid_Value_Size of value
+
   val add : t -> key -> value -> unit
 
   val iter : (key -> value -> unit) -> t -> unit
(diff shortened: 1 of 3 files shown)


src/index.ml
```diff
@@ -80,8 +82,14 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
   let entry_sizeL = Int64.of_int entry_size
 
   let append_entry io e =
-    IO.append io (K.encode e.key);
-    IO.append io (V.encode e.value)
+    let encoded_key = K.encode e.key in
+    let encoded_value = V.encode e.value in
+    if String.length encoded_key != K.encoded_size then
```
samoht: should be `<>` instead of `!=`

src/index.ml
```diff
@@ -62,6 +62,8 @@ let may f = function None -> () | Some bf -> f bf
 
 exception RO_Not_Allowed
 
+exception Different_Sizes_Not_Allowed
```
pascutto: What would you think of having exceptions like
```
exception Invalid_Key_Size of key
exception Invalid_Value_Size of value
```
to get better feedback ?

CraigFe: Seems good to me. Would also make the `check_raises` tests slightly more precise.

pascutto:
Thank you !

PR #37, opened 2019-08-12
Add close function

icristescu:
Added a close function to the IO module.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -105,8 +105,11 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
     IO.clear t.log;
     may Bloomf.clear t.entries;
     Tbl.clear t.log_mem;
-    may (fun i -> IO.clear i.io) t.index;
-    (match t.index with None -> () | Some index -> IO.close index.io);
+    may
+      (fun i ->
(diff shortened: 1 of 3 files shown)


src/index.ml
```diff
@@ -106,6 +106,7 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
     may Bloomf.clear t.entries;
     Tbl.clear t.log_mem;
     may (fun i -> IO.clear i.io) t.index;
+    (match t.index with None -> () | Some index -> IO.close index.io);
```
CraigFe: This could be pushed into the `may` function above, which is there to handle exactly this type of pattern :slightly_smiling_face:

src/index.ml
```diff
@@ -283,6 +284,7 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
       Tbl.clear t.log_mem;
       iter_io add_log_entry t.log;
       let index_path = index_path t.root in
+      (match t.index with None -> () | Some index -> IO.close index.io);
```
```suggestion
      may (fun i -> IO.close i.io) t.index;
```

pascutto:
Thanks !

PR #38, opened 2019-08-16
Some code suggestions

icristescu:
A few code suggestions

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -157,17 +157,17 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
 
   let iter_io ?min ?max f io = iter_io_off ?min ?max (fun _ e -> f e) io
 
-  type window = { buf: bytes; off: int64 }
+  type window = { buf : bytes; off : int64 }
 
   let get_entry io ~window off =
     match window with
     | None ->
-      let buf = Bytes.create entry_size in
-      let _ = IO.read io ~off buf in
-      decode_entry buf 0
+        let buf = Bytes.create entry_size in
+        let _ = IO.read io ~off buf in
+        decode_entry buf 0
     | Some w ->
-      let off = Int64.(to_int @@ sub off w.off) in
-      decode_entry w.buf off
+        let off = Int64.(to_int @@ sub off w.off) in
+        decode_entry w.buf off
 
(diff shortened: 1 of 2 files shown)


src/index.ml
```diff
@@ -181,7 +181,7 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
       if readonly then None
       else Some (Bloomf.create ~error_rate:0.01 100_000_000)
     in
-    let log_mem = Tbl.create 1024 in
+    let log_mem = Tbl.create log_size in
```
pascutto: We used to have that initialization a few weeks ago, but reduced the initial size since there could be use cases where we add fewer entries than that and don't want to allocate extra memory.

samoht: agreed, let's keep a small `log_mem` and let the OCaml runtime change it when needed.

src/index.ml
```diff
@@ -157,17 +157,17 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
-  type window = { buf: bytes; off: int64 }
+  type window = { buf : bytes; off : int64 }
 
   let get_entry io ~window off =
     match window with
     | None ->
-      let buf = Bytes.create entry_size in
-      let _ = IO.read io ~off buf in
-      decode_entry buf 0
+        let buf = Bytes.create entry_size in
```
samoht: is this using ocamlformat?

src/index.ml
```diff
@@ -157,17 +157,17 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+        let _ = IO.read io ~off buf in
```
cfcs: If `IO.read` doesn't completely fill `buf` here (hopefully `~off` is an offset into `io`?), the uninitialized memory from `Bytes.create` will be in `buf` when `decode_entry` is called.
Same for the other cases where the result is ignored.

src/index.ml
icristescu: Yes, the 0.10 version as specified by .ocamlformat...But I get the same formatting for 0.11.0

samoht: would probably make sense to update `.ocamlformat` to use 0.11.0 as well.

PR #39, opened 2019-08-26
batch reads during merges

samoht:
Before:

```
Adding 3000000 bindings.
	2999k/3000k
3000000 bindings added in 18.889829s.
...
```

After:

```
Adding 3000000 bindings.
	2999k/3000k
3000000 bindings added in 14.619115s.
...
```

e.g. ~25%  improvements of the merge speed.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -352,23 +352,32 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
   (** Merge [log] with [t] into [dst_io].
       [log] must be sorted by key hashes. *)
   let merge_with log index dst_io =
-    let buf = Bytes.create entry_size in
+    let entries = 10_000 in
+    let buf = Bytes.create (entries * entry_size) in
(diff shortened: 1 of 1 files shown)


src/index.ml
```diff
@@ -118,7 +118,7 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
 
   let index_path root = root // "index" // "index"
 
-  let page_size = Int64.mul entry_sizeL 1_000L
+  let page_size = Int64.mul entry_sizeL 1L
```
pascutto: This seems like the opposite of batch reading, doesn't it hurt the performance of the operations using `iter_io` ?

samoht: do we use `iter_io` in the benchmarks? :-)

But yes, you are right this is used only in iter_io so that should not change.

pascutto: No we don't, `iter_io` is used on startup, and to sync readonly instances.
I'll add some more benchs.

PR #40, opened 2019-08-26
batch reads during searches

samoht:
Before:

```
...
Finding 3000000 bindings.
	2999k/3000k
3000000 bindings found in 51.581994s.
...
```

After:

```
...
Finding 3000000 bindings.
	2999k/3000k
3000000 bindings found in 43.150246s.
...
```

e.g. a bit less than ~15% improvements.

We are still doing too many reads (almost 2Go for a 170Mo file!) but way less than before (it used to be 16 Go).

Note that I've also fixed the `look_around` function to always stay between `low` and `high`.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -222,7 +222,7 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
   let look_around ~window io init ~low ~high key h_key off =
     let rec search acc op curr =
       let off = op curr entry_sizeL in
-      if off < 0L || off >= IO.offset io || off < low || off >= high then acc
+      if off < 0L || off >= IO.offset io || off < low || off > high then acc
       else
         let e = get_entry ~window io off in
         let h_e = e.key_hash in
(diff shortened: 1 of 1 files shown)


src/index.ml
```diff
@@ -208,16 +215,16 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
     | Some e -> e
-    | None -> get_entry io off
+    | None -> get_entry ~window io off
 
-  let look_around io init key h_key off =
+  let look_around ~window io init ~low ~high key h_key off =
     let rec search acc op curr =
       let off = op curr entry_sizeL in
-      if off < 0L || off >= IO.offset io then acc
+      if off < 0L || off >= IO.offset io || off < low || off >= high then acc
```
pascutto: I think `high` should be included here.
```suggestion
      if off < 0L || off >= IO.offset io || off < low || off > high then acc
```

PR #43, opened 2019-08-27
Use an Array abstraction within the interpolation search

CraigFe:
This provides the interpolation_search function with a cleaner abstraction of the underlying index as an array of elements linearly distributed according to some 'metric' (in this case, the key hash).

This simplifies the core of the interpolation search quite a bit, and allows it to be tested individually. Unfortunately, it needs rebasing over https://github.com/mirage/index/pull/39 which breaks the clean array abstraction I was going for.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -166,19 +166,6 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
 
   let iter_io ?min ?max f io = iter_io_off ?min ?max (fun _ e -> f e) io
 
-  type window = { buf : bytes; off : int64 }
-
-  let get_entry io ~window off =
-    match window with
-    | None ->
-        let buf = Bytes.create entry_size in
-        let n = IO.read io ~off buf in
-        assert (n = entry_size);
-        decode_entry buf 0
-    | Some w ->
-        let off = Int64.(to_int @@ sub off w.off) in
-        decode_entry w.buf off
-
   module Entry = struct
     type t = entry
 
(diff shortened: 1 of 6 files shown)


samoht:
squashed and rebased -- there are new errors in the tests but it's a bit hard to debug right now as running them individually works fine!

src/index.ml
```diff
@@ -167,18 +170,151 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+            e
+          with _ -> assert false )
+      | _ -> get_entry_from_io t.io off
+
+    let length t = Int64.(div (IO.offset t.io) entry_sizeL)
+
+    let set_buffer t ~low ~high ~range =
+      let buf = Bytes.create range in
+      assert (
+        let n = IO.read t.io ~off:low buf in
```
pascutto: After a debugging session with @CraigFe: 
`low` is an index in the array abstraction here, while we would expect and offset, multiplying by `entry_size` is necessary.

CraigFe: Resolved by https://github.com/mirage/index/pull/43/commits/114b1a3c002db426dae61f1830f22a7b1d438bd3. Thanks for your help with debugging this.

pascutto:
https://github.com/mirage/index/pull/43/commits/c9601a0aa5d28b4f8e58a84093d152f2e81ab4e4 has fixed the performance drop, I can see benchmarks similar to master now :+1: .

PR #49, opened 2019-08-28
[WIP] Add metrics

CraigFe:
Eventually we'd like to track these metrics:
 - [x] number of reads
 - [ ] number of writes
 - [x] bytes read
 - [ ] bytes written
 - [ ] width of fan-out range
 - [ ] general assurance that the interpolation search is performant (e.g. steps / search distribution)
 - [x] GC stats

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -71,6 +71,8 @@ let may f = function None -> () | Some bf -> f bf
 exception RO_not_allowed
 
 module Make (K : Key) (V : Value) (IO : IO) = struct
+  module IO = Index_metrics.Instrument_IO (IO)
+
   type key = K.t
 
   type value = V.t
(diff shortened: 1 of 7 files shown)


src/index.ml
```diff
@@ -66,7 +66,49 @@ let src = Logs.Src.create "index" ~doc:"Index"
 
 module Log = (val Logs.src_log src : Logs.LOG)
 
+module IndexMetrics = struct
```
samoht: that case is a not conventional in OCaml lands :-) `Index_metrics` ?

CraigFe: Indeed :wink: I actually don't like this being in the same file at all; I could define these metrics in a separate `index_metrics.ml` unit, along with the IO instrementation functor. Thoughts?

samoht: sounds better indeed

CraigFe:
Closing as this has since been superceded by @icristescu's work.

PR #52, opened 2019-08-28
Expose merge for testing

icristescu:
Expose a force merge in the API and add tests

### DIFF
--- a/test/unix/force_merge.ml
+++ b/test/unix/force_merge.ml
@@ -1,0 +1,218 @@
+let reporter ?(prefix = "") () =
+  let report src level ~over k msgf =
+    let k _ =
+      over ();
+      k ()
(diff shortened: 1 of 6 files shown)


test/unix/force_merge.ml
```diff
@@ -0,0 +1,218 @@
...
+      Index.flush t;
+      tbl )
+    else
+      let k = Key.v () in
+      let v = Value.v () in
+      Index.replace t k v;
+      Hashtbl.add tbl k v;
+      loop (i - 1)
+  in
+  loop index_size
```
samoht: To avoid code duplication, the prelude of that file should probably be refactored in a different `common.ml` or `prelude.ml` file and be shared between all the tests.

PR #53, opened 2019-08-29
Lock file for merge

icristescu:
Lock mechanism with a index.lock file for merging. 

Thanks to @CraigFe for remarking that dune runtests are concurrent.


CraigFe:
I'm slightly confused here; is the idea to rely on `runtest` concurrency to test the lock feature?

pascutto:
I think the presence of `tmp/index` is a lock in itself, since this is present iff a merge is in progress (or the program has crashed in the middle of a lock).
Btw if the program crashes in the middle of a merge, the `lock` file would remain and block any further execution, right ?

icristescu:
> I'm slightly confused here; is the idea to rely on `runtest` concurrency to test the lock feature?

Well, yes. 
When running the tests, it complains because it tries to rename a non-existing tmp file. Which  it suggests two concurrent ongoing merges, no ?

src/index.ml
```diff
@@ -430,6 +430,7 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
   let merge ~witness t =
     Log.debug (fun l -> l "merge %S" t.root);
     let tmp_path = t.root // "tmp" // "index" in
+    IO.wait_for_lock tmp_path;
```
samoht: I am not sure to understand why do we wait for the path to disappear?

Would probably be cleaner to use `flock` like @mefyl did on [dune cache](https://github.com/ocaml/dune/pull/2493/commits/7cf3776981ff0c965856793130705892f3b5d8db) to ensure there is only 1 writer process.

src/io.mli
```diff
@@ -30,4 +30,10 @@ module type S = sig
   val close : t -> unit
 
   val valid_fd : t -> bool
+
+  type fd
+
+  val lock : string -> fd
```
CraigFe: w.r.t making this IO signature Mirage-compatible in future, I wonder if something like:
```ocaml
val with_lock: (unit -> 'a) -> 'a
```
might be a less Unix-ish abstraction. Plus, manual lock handling makes me nervous. (any thoughts, @dinosaure ?)

CraigFe: (obviously, the manual `lock`/`unlock` functions are perfectly OK. Just conjecturing)

samoht: I think in that case it's a bit difficult to do differently as the lock has to hold for the duration of the program.

pascutto: Besides this, the type name doesn't need to leak information about the implementation, maybe `type lock` would be easier to understand in the API ?

src/unix/index_unix.ml
```diff
@@ -254,6 +254,34 @@ module IO : Index.IO = struct
...
+    with
+    | Unix.Unix_error (Unix.EAGAIN, _, _) ->
+        Unix.close fd;
+        None
+    | e ->
+        Unix.close fd;
+        raise e
+
+  let lock path =
+    match _lock Unix.F_LOCK path with
```
samoht: maybe use `F_RLOCK` to have a proper exception?

src/unix/index_unix.ml
```diff
@@ -254,6 +254,34 @@ module IO : Index.IO = struct
...
+    | Some fd -> fd
+    | None -> failwith "lockf F_LOCK didn't lock"
```
samoht: maybe use a more informative error message, including the name of the lock file?

src/unix/index_unix.ml
icristescu: Then both tests cannot run, one of them raises the exception. 

Do I remove the concurrent test ?

src/io.mli
dinosaure: I can confirm that a `type lock` should help a lot for a Mirage compat', like:
```ocaml
type 'a io
type lock

val create : unit -> lock
val lock : lock -> unit io
val unlock : lock -> unit
```

src/unix/index_unix.ml
samoht: yes, we should have a deterministic way to test this. For instance, use `Unix.exec` with a process trying to open that file and check the exit code.

For an end-user perspective, the role of that lock file is to give a feedback quickly that the API is used wrongly (e.g. there *must* be only one writer process). Blocking will not really help here :-)

icristescu: For the moment I remove the concurrent test altogether, so that you can merge this if you want. I'll add the concurrent tests once I figured out how to use `Unix.exec`.

src/unix/index_unix.ml
```diff
@@ -260,6 +260,34 @@ module IO : Index.IO = struct
       let _ = Unix.fstat t.raw.fd in
       true
     with Unix.Unix_error (Unix.EBADF, _, _) -> false
+
+  type lock = Unix.file_descr
+
+  let _lock op f =
```
pascutto: To be consistent with the rest of the module (and avoid ignored warning `32`), could you rename this to `unsafe_lock` for instance ?

PR #55, opened 2019-08-29
Read after clear does not fail

icristescu:
Adding a test

### DIFF
--- a/test/unix/main.ml
+++ b/test/unix/main.ml
@@ -189,26 +189,12 @@ let readonly_clear () =
   Index.flush w;
   test_find_present r;
   Index.clear w;
-  Index.flush w;
   Hashtbl.iter
     (fun k _ ->
       match Index.find r k with
       | _ -> Alcotest.fail (Printf.sprintf "Found %s key after clearing." k)
       | exception Not_found -> ())
     tbl
-
-(*ignore (Index.mem r (Key.v ()));
-  Index.clear w;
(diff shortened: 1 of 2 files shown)


test/unix/main.ml
```diff
@@ -182,6 +182,34 @@ let readonly () =
...
+  Index.clear w;
+  Index.flush w;
+  Hashtbl.iter
+    (fun k _ ->
+      match Index.find r k with
+      | _ -> Alcotest.fail (Printf.sprintf "Found %s key after clearing." k)
+      | exception Not_found -> ())
+    tbl
+
+(*ignore (Index.mem r (Key.v ()));
```
samoht: Should these comments be removed before merging that PR?

test/unix/main.ml
```diff
@@ -182,6 +182,20 @@ let readonly () =
...
+  let r = Index.v ~fresh:false ~readonly:true ~log_size "index1" in
+  Hashtbl.iter (fun k v -> Index.replace w k v) tbl;
+  Index.flush w;
+  test_find_present r;
+  Index.clear w;
+  Hashtbl.iter
+    (fun k _ ->
+      match Index.find r k with
+      | _ -> Alcotest.fail (Printf.sprintf "Found %s key after clearing." k)
+      | exception Not_found -> ())
```
pascutto: We should use `Alcotest.check_raises` here instead of pattern-matching.

PR #63, opened 2019-08-30
it's actually fine to create a file in read-only mode

samoht:


samoht:
I guess we didn't that before, so the tests keep working :p

pascutto:
I think there was a test for that in #33

PR #66, opened 2019-09-02
Concurrent tests

icristescu:


CraigFe:
I think this test should be avoided on Windows using something like [this](https://github.com/mirage/irmin/blob/858e998fc4bd159e68053dc5da476f8e018fa6fd/test/irmin-http/test_http.ml#L157). 

Also, we should probably add a Windows CI runner so that such failures are obvious.

test/unix/concurrent.ml
```diff
@@ -0,0 +1,29 @@
...
+let concurrent_runs () =
+  if Sys.os_type = "Win32" then ()
+  else
+    let exn =
+      Failure
+        ("Lock didn't succeed: file " ^ index_name ^ "/index/.lock is present")
+    in
+    let pid = Unix.fork () in
+    if pid <> 0 then (
+      Unix.sleepf 0.001;
```
samoht: this might be a bit fragile, but I guess that's good enough for now. We can restructure this later if we see that this the tests is not deterministic enough.

CraigFe: Now that we have a `Private` functor abstraction, could we add a hook for the creation of the lock file to test this more consistently?

It would be nice to have these tests merged so that they're not a maintenance burden to you :slightly_smiling_face:

icristescu: Thanks for the idea, I removed all calls to `Unix.sleep`.

CraigFe:
I'm looking into the spurious test failures here. Will start with a rebase.

samoht:
Keeping this open as this seems to be a real synchronisation issue and need more investigation.

samoht:
@CraigFe is planning to have a look at it at one point :-)

PR #73, opened 2019-09-03
Add README

### LINKED ISSUE #68: Add a README

CraigFe:
This resolves #68.

Have kept the README fairly sparse for now, since the library is reasonably WIP. I suggest adding more details (documentation pointers, etc.) once the IO signature is a little more stable.

### DIFF
--- a/src/index.mli
+++ b/src/index.mli
@@ -1,15 +1,18 @@
-(** Deudex
+(** Index
 
-    deudex is a scalable implementation of persistent indexes in Ocaml.
+    [Index] is a scalable implementation of persistent indices in OCaml.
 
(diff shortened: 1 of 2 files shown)


pascutto:
Thanks for fixing the messy documentation !

src/index.mli
```diff
@@ -28,23 +31,23 @@ module type Key = sig
   (** The maximum number of bits used to encode hashes. `Hashtbl.hash` uses 30
       bits. *)
 
+  val encoded_size : int
```
samoht: this breaks the build

CraigFe: Interesting. Regardless, it's not necessary to change the order here. Reverted.

PR #75, opened 2019-09-03
Add checks on file descriptors

icristescu:


test/unix/dune
```diff
@@ -1,3 +1,3 @@
 (tests
  (names main force_merge)
- (libraries index.unix alcotest fmt logs logs.fmt))
+ (libraries index.unix alcotest fmt logs logs.fmt str))
```
CraigFe: I think the `Str` module is a bit unpopular (https://github.com/mirage/mirage/issues/663). Since it's only a test dependency, it maybe doesn't matter that much, but I think it's generally to be avoided.

icristescu: Thanks, I didn't know. I'll take a look at Astring.

samoht: it's unpopular because it is not thread safe (and the API is unsafe) ;-)

samoht: for regular expression matching, it's probably better to use `Re`

test/unix/force_merge.ml
```diff
@@ -12,20 +12,58 @@ let test_find_present t =
...
+      if not (res = v) then Alcotest.fail "Replacing existing value failed."
   | exception Not_found ->
-      Alcotest.fail
-        (Printf.sprintf "Inserted value is not present anymore: %s." k)
+      Alcotest.failf "Inserted value is not present anymore: %s." k
+
+let test_fd () =
+  let pid = string_of_int (Unix.getpid ()) in
+  let fd_file = "tmp" in
+  let lsof_command = "lsof -a -s -p " ^ pid ^ " > " ^ fd_file in
```
CraigFe: Not all Unix systems have `lsof`; mine didn't. The tests probably shouldn't be failing on such systems; would be better to just skip the test altogether.

PR #79, opened 2019-09-09
Concurrent read

icristescu:


bench/concurrent_bench.ml
```diff
@@ -0,0 +1,19 @@
+open Main
+
+let read_index () =
+  let r = Index.v ~fresh:false ~readonly:true ~log_size index_name in
+  Fmt.epr "\n Read in readonly index. \n";
+  let count = ref 0 in
+  Index.iter
```
samoht: does this work when the index is updated concurrently? e.g. will new written elements be read too?

pascutto: Not for readonly instances. Fixed by #81 .

pascutto:
Can this bench be integrated to `main.ml` ? I feel like the file is short enough for now to avoid having to deal with multiple files/`open`.
Or if we really want to split, we can think of a better architecture (and apply it to the tests too)

icristescu:
> Can this bench be integrated to `main.ml` ? I feel like the file is short enough for now to avoid having to deal with multiple files/`open`.

I can integrate it but I'm not sure we should, as it does not really belong to benchmarking. It is actually a test that uses the heavy adds and reads of bench. 

> Or if we really want to split, we can think of a better architecture (and apply it to the tests too)

I think it is a bit difficult to copy the architecture of Irmin tests for example, because here we have a lot of tests that use fork. Whenever there is a fork all tests that follow are run twice, so we have to be careful to separate the tests properly, and I haven't yet figured out a way to do it.

bench/concurrent_bench.ml
pascutto: Oh I think I understand what you meant now.
The `iter` indeed won't see the elements added after it was called., so this bench will only read the elements that were there from previous benchs.

icristescu:
These tests are less complete than the ones in PR #83 so I'm closing this PR.

PR #80, opened 2019-09-10
add benchmarks for read-only Index.find

samoht:


pascutto:
Looks good to me.
Would be nice if the RO instance had a different name than the RW one, eg `r` and `w`, to avoid confusion in the future.

samoht:
> Would be nice if the RO instance had a different name than the RW one
fixed

PR #83, opened 2019-09-10
Testing concurrent reads and writes

icristescu:


test/unix/concurrent_ro.ml
```diff
@@ -0,0 +1,100 @@
+open Common
+
+let index_name = Filename.concat "_tests" "unix.concurrent_ro"
+
+let tbl = tbl index_name
+
+let w = Index.v ~fresh:false ~log_size index_name
```
CraigFe: I'd prefer constructing this inside the test (with a subscoped directory name), along with the pipes below, and passing them explicitly to the helper functions that need them. 

These sorts of global test instances seem to lead to a mess :slightly_smiling_face:

icristescu: Done, but for now, it still depends on the index being populated by the tbl in common. 

`index_name` is defined above and subscoped to `_tests`. Is this what you meant?

CraigFe: If this test uses `index_name` to construct it's instance-under-test, future tests will also test the same instance on disk. I was suggesting doing something like [this](https://github.com/mirage/index/blob/master/test/unix/main.ml#L105) to make sure each test uses a unique name.

It's not a big deal, however, since my PR will clean up these sorts of details.

CraigFe: And w.r.t. the table being populated by Common; I'm also working on that :slightly_smiling_face:

PR #84, opened 2019-09-10
Refactor tests

CraigFe:
Makes many of the Unix tests individually executable.

### DIFF
--- a/test/unix/common.ml
+++ b/test/unix/common.ml
@@ -23,8 +23,6 @@ let report () =
 
 let string_size = 20
 
-let index_size = 103
-
 let () = Random.self_init ()
 
 let random_char () = char_of_int (33 + Random.int 94)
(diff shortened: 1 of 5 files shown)


test/unix/common.ml
```diff
@@ -67,26 +65,42 @@ end
...
+  let c = ref 0
+
+  let name_of_count () =
+    Filename.concat Config.root ("index_" ^ string_of_int !c)
+
+  (* Fresh, empty index. *)
+  let empty_index () =
+    incr c;
+    let name = name_of_count () in
+    Fmt.pr "Constructing empty_index context object: %s" name;
```
samoht: Use `Log.info` instead of `Fmt.pr`?

CraigFe:
This currently weakens some of the testing in `force_merge`... working on it!

CraigFe:
In fact, under further investigation, it looks like these tests were originally not working 100%. Have added a TODO here: https://github.com/mirage/index/pull/84/files#diff-81aec069a46f58216cc2e7046f8d183dR28.

test/unix/main.ml
```diff
@@ -4,234 +4,242 @@ let ( // ) = Filename.concat
...
+    let exn = Unix.Unix_error (Unix.EBADF, "read", "") in
+    Alcotest.check_raises "Cannot read in readonly index after close." exn
+      (fun () -> ignore (Index.find t k))
+
+  let fail_read_after_close () =
+    let w = Index.v ~fresh:true ~readonly:false ~log_size (root // "test4") in
+    Hashtbl.iter (fun k v -> Index.replace w k v) main.tbl;
+    let k = Key.v () in
+    let v = Value.v () in
+    Index.replace w k v;
```
icristescu: You need to add the pair `k,v` to the hash table too.

CraigFe: Good catch; we should iterate in both directions to catch this.

CraigFe:
Have now integrated @icristescu's suggestions and strengthened the tests to check for two-way equivalence of the indices and hashtables (before we were just checking that the hashtable was a subset of the index).

test/unix/common.ml
```diff
@@ -67,26 +65,44 @@ end
...
-      tbl )
-    else
+module Make_context (Config : sig
+  val root : string
+end) =
+struct
+  let c = ref 0
+
+  let name_of_count () =
+    Filename.concat Config.root ("index_" ^ string_of_int !c)
```
pascutto: Maybe this would be nicer with something like this:
```ocaml
let new_name =
  let c = ref 0 in 
  fun () ->
    let n = Filename.concat Config.root ("index_" ^ string_of_int !c) in
    incr c;
    n
```

CraigFe: Agreed. I've also pushed the log entry for the new name into this function too. Let me know what you think :slightly_smiling_face:

pascutto: LGTM :+1:

PR #85, opened 2019-09-10
Bug fix: Fan import should not use encoded_size

icristescu:


src/index.ml
```diff
@@ -289,9 +289,7 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
         let io =
           IO.v ~fresh:false ~readonly:true ~generation ~fan_size:0L index_path
         in
-        let fan_out =
-          Fan.import ~hash_size:K.encoded_size (IO.get_fanout io)
-        in
+        let fan_out = Fan.import ~hash_size:K.hash_size (IO.get_fanout io) in
```
samoht: as a bonus of being a nice fix, it also makes the expression fits on one line 🥇

pascutto:
Good catch ! :100:

PR #91, opened 2019-09-17
Benchmarks

icristescu:
Similar to the ones in LevelDB and RocksDB.

### DIFF
--- a/bench/common.ml
+++ b/bench/common.ml
@@ -1,0 +1,76 @@
+let reporter ?(prefix = "") () =
+  let report src level ~over k msgf =
+    let k _ =
+      over ();
+      k ()
+    in
+    let ppf = match level with Logs.App -> Fmt.stdout | _ -> Fmt.stderr in
+    let with_stamp h _tags k fmt =
+      let dt = Unix.gettimeofday () in
+      Fmt.kpf k ppf
+        ("%s%+04.0fus %a %a @[" ^^ fmt ^^ "@]@.")
+        prefix dt
+        Fmt.(styled `Magenta string)
+        (Logs.Src.name src) Logs_fmt.pp_header (level, h)
+    in
+    msgf @@ fun ?header ?tags fmt -> with_stamp header tags k fmt
+  in
+  { Logs.report }
+
+let report () =
+  Logs.set_level (Some Logs.Info);
+  Logs.set_reporter (reporter ())
+
+let random_char () = char_of_int (33 + Random.int 94)
+
+let random_string string_size =
+  String.init string_size (fun _i -> random_char ())
+
+module Make_context (Config : sig
+  val key_size : int
+
+  val hash_size : int
+
+  val value_size : int
+end) =
+struct
+  module Key = struct
+    type t = string
+
+    let v () = random_string Config.key_size
+
+    let hash = Hashtbl.hash
+
+    let hash_size = Config.hash_size
+
+    let encode s = s
+
+    let decode s off = String.sub s off Config.key_size
+
+    let encoded_size = Config.key_size
+
+    let equal = String.equal
+
+    let pp s = Fmt.fmt "%s" s
+  end
+
+  module Value = struct
+    type t = string
+
+    let v () = random_string Config.value_size
+
+    let encode s = s
+
+    let decode s off = String.sub s off Config.value_size
+
+    let encoded_size = Config.value_size
+
+    let pp s = Fmt.fmt "%s" s
+  end
+end
+
+let with_timer f =
+  let t0 = Sys.time () in
+  let x = f () in
+  let t1 = Sys.time () -. t0 in
+  (x, t1)
--- a/bench/common.mli
+++ b/bench/common.mli
@@ -1,0 +1,25 @@
+val report : unit -> unit
+(** Set logs reporter at [Logs.Info] level *)
+
+module Make_context (Config : sig
+  val key_size : int
+
+  val hash_size : int
+
+  val value_size : int
+end) : sig
+  module Key : sig
+    include Index.Key with type t = string
+
+    val v : unit -> t
+  end
+
+  module Value : sig
+    include Index.Value with type t = string
+
+    val v : unit -> t
+  end
+end
+
+val with_timer : (unit -> 'a) -> 'a * float
+(** [with_timer f] returns the execution time of f in seconds *)
--- a/bench/db_bench.ml
+++ b/bench/db_bench.ml
@@ -13,15 +13,13 @@ let () = Random.init seed
 
 let key_size = 44
 
-(* the hash_size in pack_index*)
-let hash_size = 60
+let hash_size = 30
 
 let value_size = 13
 
 let nb_entries = 1_000_000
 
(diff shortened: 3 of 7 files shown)


bench/dune
```diff
@@ -1,6 +1,13 @@
...
  (name main)
- (libraries fmt index.unix))
+ (modules main)
+ (libraries fmt index.unix common))
+
+(executable
+ (name db_bench)
+ (modules db_bench)
+ (libraries fmt index.unix common lmdb bigstring rresult))
+
```
CraigFe: This extra line causes `dune build @fmt` to fail.

bench/db_bench.ml
```diff
@@ -0,0 +1,267 @@
...
+  let read_random r = print_results (read r random) nb_entries
+
+  let read_seq r = print_results (read r sorted) nb_entries
+
+  let read_seq_rev r = print_results (read r (List.rev sorted)) nb_entries
+
+  let close env = closedir env
+end
+
+let () =
```
CraigFe: I think it's worth putting this in a `main` function and then having `let () = main ()` at the top-level, in case we want to invoke this without calling the executable.

bench/db_bench.ml
```diff
@@ -0,0 +1,267 @@
...
+  Index.init ();
+  Lmdb.cleanup ();
+  Fmt.epr "Keys: %d bytes each.\n%!" key_size;
```
CraigFe: Perhaps we should be using `Logs` rather than `Fmt.epr` here?

bench/common.mli
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  module Value : sig
+    include Index.Value with type t = string
+
+    val v : unit -> t
+  end
+end
+
+val with_timer : (unit -> 'a) -> 'a * float
```
CraigFe: Potentially warrants a comment saying that the returned time is in units of seconds.

bench/common.ml
```diff
@@ -0,0 +1,76 @@
+let reporter ?(prefix = "") () =
```
samoht: don't forget the copyright headers when you create a new file :-)

bench/db_bench.ml
```diff
@@ -165,11 +184,11 @@ module Lmdb = struct
   let fail_on_error f =
     match f () with Ok _ -> () | Error err -> failwith (string_of_error err)
 
-  let flags = [ Lmdb.NoRdAhead; Lmdb.NoSync ]
+  let flags = [ Lmdb.NoRdAhead; Lmdb.NoSync; Lmdb.NoSync ]
```
samoht: duplicated `NoSync`?
for the sequential reads tests it might make sense to remove the `NoRdAhead` flag as it will help the OS to pre-fetch sequential pages.

bench/db_bench.ml
```diff
@@ -221,9 +240,20 @@ module Lmdb = struct
...
-  let read_seq r = print_results (read r sorted) nb_entries
-
-  let read_seq_rev r = print_results (read r (List.rev sorted)) nb_entries
+  let read_seq r =
+    let read (txn, ddb) () =
+      opencursor txn ddb >>= fun cursor ->
+      cursor_first cursor >>= fun () ->
+      cursor_iter
+        ~f:(fun (k, v) ->
+          ignore (Bigstring.to_string k);
```
samoht: you probably don't want to call `Bigstring.to_string` at all here, as this is copying bytes and allocating a string in the heap which is not what we want to measure -- I think it's just fine to ignore `k` and `v`.

src/unix/index_unix.ml
```diff
@@ -2,6 +2,18 @@ exception RO_not_allowed
 
 let current_version = "00000001"
 
+type stats = { mutable bytes_read : int; mutable nb_reads : int }
```
samoht: can you do this for writes too? that'll help computing a more precise "write amplification" result in the benchmarks.

src/unix/index_unix.mli
```diff
@@ -5,3 +5,7 @@ module Make (K : Index.Key) (V : Index.Value) :
 module Private : sig
   module IO : Index.IO
 end
+
+val reset_stats : unit -> unit
+
+val get_stats : unit -> int * int
```
samoht: probably best to just exposes the `stats` type.

bench/db_bench.ml
icristescu: > duplicated `NoSync`?

typo, it was supposed to be ` Lmdb.NoMetaSync` 

> for the sequential reads tests it might make sense to remove the `NoRdAhead` flag as it will help the OS to pre-fetch sequential pages.

bench/db_bench.ml
icristescu: don't we want to add the time to convert the big string to the overall time ?

samoht: I guess it really depends on what we are trying to measure :-)

bench/common.ml
icristescu: there are no copyright headers in any file in the index

samoht: good point! happy to have a fix in a separate PR

bench/db_bench.ml
```diff
@@ -14,13 +14,13 @@ let () = Random.init seed
...
-let hash_size = 60
+let hash_size = 30
 
 let value_size = 13
 
 let nb_entries = 1_000_000
 
-(* in the tezos use case, log_size is proportional to the nb_entries *)
+(* in tezos the log_size is 500_000  *)
 let log_size = nb_entries / 5
```
samoht: can't we use `log_size = 500_000` here too?

icristescu: We can, but I think we are a bit lying to ourselves. The larger the log size the better we do for these small benchmarks. But then what we measure here will no reflect the performance for a bigger number of entries. 
Maybe it's fine, as probably lmdb's perf degrades also for large datasets.

bench/db_bench.ml
```diff
@@ -0,0 +1,329 @@
...
+         file_size = %f%!"
+        ratio_bytes ratio_reads ratio_file_size)
+
+let read_amplif nb_entries =
+  let stats = Index_unix.get_stats () in
+  let ratio_bytes =
+    float_of_int stats.bytes_read /. float_of_int (entry_size * nb_entries)
+  in
+  let ratio_reads = float_of_int stats.nb_reads /. float_of_int nb_entries in
+  Log.info (fun l ->
```
samoht: `Log.app` is probably better here to avoid cluttering the output with debug prefixes.

bench/db_bench.ml
samoht: we are not lying, we are just using the same cache parameters as used in production. We should make sure we are not introducing bias so for instance we should make sure that the read tests are done on "cold" boot, e.g. with empty caches. But it's perfectly fine to use and have the right write caches. lmdb is doing the same internally!

samoht:
Great work! I think this is good enough the be merged and future improvement can come in new PRs. I'm opening #93 to keep track of general improvements that I have in mind.

PR #96, opened 2019-09-20
Add index-bench.opam file

### LINKED ISSUE #95: Add an 'index-bench.opam' file to specify benchmark dependencies

CraigFe:
Resolves #95.

### DIFF
--- a/index-bench.opam
+++ b/index-bench.opam
@@ -1,0 +1,30 @@
+opam-version: "2.0"
+maintainer:   "Clement Pascutto"
+authors:      ["Clement Pascutto" "Thomas Gazagnaire" "Ioana Cristescu"]
+license:      "MIT"
+homepage:     "https://github.com/mirage/index"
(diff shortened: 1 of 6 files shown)


index-bench.opam
```diff
@@ -0,0 +1,30 @@
...
+  "index"
+  "fmt"
+  "logs"
+  "bigstring"
+  "rresult"
+  "lmdb"
+]
+
+pin-depends: [
+  [ "lmdb.0.1" "git+https://github.com/vbmithr/ocaml-lmdb" ]
```
pascutto: This is probably better since lmdb was not released
```suggestion
  [ "lmdb.dev" "git+https://github.com/vbmithr/ocaml-lmdb" ]
```

CraigFe: Agreed; I think we should also pin to a specific commit hash for maximum reproducibility.

PR #98, opened 2019-09-20
Refactor benchs

icristescu:
Replaced list with array: this decreases the execution time of random
read and write to match the performance of bench/main.ml. However, after
sorting the array, the execution time increases significantly, and it
might not be due to index. Therefore the benchs using an ordered array
should not be compared with the ones on random.
Other changes:
removed file size; print stats for lmdb (in case we want to use it);
print nb of operations / seconds.

### DIFF
--- a/bench/main.ml
+++ b/bench/main.ml
@@ -1,10 +1,10 @@
 open Common
 
-let key_size = 44
+let key_size = 32
 
-let value_size = 12
+let value_size = 13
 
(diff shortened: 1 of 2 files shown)


bench/main.ml
```diff
@@ -2,17 +2,17 @@ open Common
 
 let key_size = 44
```
samoht: I think we are using `32` in `db_bench`

PR #100, opened 2019-10-03
Metrics

icristescu:


icristescu:
For now, we can generate offline graphs with metrics. 

InfluxDB makes sense for an online instrumentation, but it requires lwt threads for all reads/writes, and this changes significantly the code. So for now, I think we can stick to metrics.

bench/db_bench.ml
```diff
@@ -17,7 +17,7 @@ let hash_size = 30
 
 let value_size = 13
 
-let nb_entries = 10_000_000
+let nb_entries = 100_000
```
samoht: Why do you change that variable in this PR?

src/unix/index_unix.ml
```diff
@@ -82,13 +109,15 @@ module IO : Index.IO = struct
       really_write t.fd off buf;
       t.cursor <- off ++ Int64.of_int (Bytes.length buf);
       stats.bytes_written <- stats.bytes_written + Bytes.length buf;
-      stats.nb_writes <- succ stats.nb_writes
+      stats.nb_writes <- succ stats.nb_writes;
+      Metrics.add src no_tags (fun m -> m stats)
```
samoht: I don't think we should mix `Metrics` and `stats` updates -- it's a complex subject but I think a good architecture would be to:
- expose more information in `stats` as you did
- add metrics and reporters at the call site (e.g. in the benchmarks?) where metrics API calls are done with the contents of `stats`.

It will be a bit less precise, but we can control the amount of things reported more easily and it will not mix various concerns in the core library. What do you think?

src/unix/index_unix.ml
```diff
@@ -359,6 +388,10 @@ module IO : Index.IO = struct
     | None -> err_rw_lock path
 
   let unlock fd = Unix.close fd
+
+  let incr_nb_merge () = stats.nb_merge <- succ stats.nb_merge
```
samoht: as the merges are not done by this file, `nb_merge` and `nb_replace` should probably be moved into a new `stats` value defined in metrics.ml.

Not that `Metric_unix.ml` can include the contents of that new stats value too.

bench/db_bench.ml
icristescu: By mistake, fixed.

src/unix/index_unix.ml
icristescu: Then we can't tell how many bytes are read/ written per io read /write, but maybe it's ok.

samoht: Yes I agree that we might loose some accuracy, but in that case the final caller can do `bytes_written / nb_writes` to get a reasonable average estimate. But you are right that it might be interesting o see a proper distribution instead of an average... I think that can be done later.

index.opam
```diff
@@ -17,6 +17,8 @@ depends: [
   "dune"    {>= "1.11.0"}
   "fmt"
   "logs"
+  "metrics"
```
samoht: would be nice to not depend on `metrics` at all in the index file (or if really needed, only as a test dependency)

samoht:
Sorry to nitpick but a few more comments:

- could you push the changes to the lmdb benchmarks in a new branch. The patch seems good, and can be merged without too much thinking :-)
- what do we gain by having a new `metrics` subpackage? It would probably be simpler (from a user point of view) to fold everything into the base `Index` package, either by having a `Metrics` sub-package or just putting the `stats` types at the toplevel.

icristescu:
> Sorry to nitpick but a few more comments:

No problem :)
> * could you push the changes to the lmdb benchmarks in a new branch. The patch seems good, and can be merged without too much thinking :-)

Done in PR https://github.com/mirage/index/pull/117
> * what do we gain by having a new `metrics` subpackage? It would probably be simpler (from a user point of view) to fold everything into the base `Index` package, either by having a `Metrics` sub-package or just putting the `stats` types at the toplevel.

I'm not sure what you mean by folding into the base package. 

I want to add stats from both `src/unix/index_unix.ml` and `src/index.ml`, so there is a common `index_metrics.ml`. But since they are in different folders, I don't know how to make it visible without making it a library.

icristescu:
> > I want to add stats from both src/unix/index_unix.ml and src/index.ml, so there is a common index_metrics.ml. But since they are in different folders, I don't know how to make it visible without making it a library.
> 
> Maybe just define a `Metrics` submodule in index.ml/index.mli?

Thanks, it worked.

src/index.mli
```diff
@@ -127,6 +127,30 @@ end
...
+    mutable nb_reads : int;
+    mutable bytes_written : int;
+    mutable nb_writes : int;
+    mutable nb_merge : int;
+    mutable nb_replace : int;
+  }
+
+  val reset_stats : unit -> unit
+
+  val get_stats : unit -> t
```
samoht: `Stats.get` reads probably better than `Stats.get_stats`

src/index.ml
```diff
@@ -4,6 +4,57 @@ module Private = struct
   module Search = Search
 end
 
+module Stats = struct
```
CraigFe: Did you consider defining this in `stats.ml`?

icristescu: I tried :) but I would like to call it from two files in different folders (`index.ml` and `index_unix.ml`) and avoid making it a library, and I didn't managed to do it otherwise.

samoht: as there is in an `index.mli` the `Stats` module will be hidden by default. You need to add `module Stats = Stats` to export it in the public interface.

icristescu: Thanks, it worked !

PR #102, opened 2019-10-04
Cleaner `force_merge` signature

pascutto:
Add a way to get the witness from the index so that it's not required in the `force_merge` signature.
`get_witness` is currently not very nice so review is appreciated.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -63,7 +63,7 @@ module type S = sig
 
   val close : t -> unit
 
-  val force_merge : t -> key -> value -> unit
+  val force_merge : t -> unit
 end
 
 let may f = function None -> () | Some bf -> f bf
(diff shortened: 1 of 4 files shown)


src/index.ml
```diff
@@ -422,9 +422,23 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
         IO.set_generation t.log generation;
         t.generation <- generation
 
-  let force_merge t key value =
+  exception Found of entry
```
CraigFe: Consider defining this exception in `get_witness` with
```ocaml
  let get_witness t =
    let exception Found of entry in
```
to limit the damage.

src/index.ml
```diff
@@ -422,9 +422,22 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+    try
+      Tbl.iter (fun _ entry -> raise (Found entry)) t.log_mem;
+      match t.index with
+      | None -> None
+      | Some index ->
+          let buf = Bytes.create entry_size in
+          let n = IO.read index.io ~off:0L buf in
+          assert (n = entry_size);
+          Some (decode_entry buf 0)
+    with Found e -> Some e
```
Julow: You could `try` only the `Tbl.iter ` line with `match Tbl.iter .. with exception Found -> ... | () -> ...` to avoid having unrelated code in the `try` block.

PR #105, opened 2019-10-08
Review the use of logs

pascutto:
The goal is to have more logs, in a more consistent way.

### DIFF
--- a/src/search.ml
+++ b/src/search.ml
@@ -52,12 +52,7 @@ module type S = sig
   module Array : ARRAY with type elt = Entry.t
 
   val interpolation_search :
-    root:string ->
-    Array.t ->
-    Entry.Key.t ->
-    low:int64 ->
-    high:int64 ->
-    Entry.Value.t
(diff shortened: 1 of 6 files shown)


pascutto:
I think we can consider merging this, and adding more later if needed.

src/search.ml
```diff
@@ -97,25 +110,24 @@ module Make
 
   (** Improves over binary search in cases where the values in some array are
       uniformly distributed according to some metric (such as a hash). *)
-  let interpolation_search array key ~low ~high =
+  let interpolation_search ~root array key ~low ~high =
```
CraigFe: I'm bit uneasy about pushing the `root` value in here unless it's really necessary; it effectively gives this portion of the code ability to violate the array semantics and, in particular, requires passing unhelpful values in the case where no IO is being used (https://github.com/mirage/index/pull/105/files#diff-211cfa6449d8c6514e18929867251ec2R65). It seems that it's only used to report either `found` or `not_found`, which I think could be done at a higher level?

CraigFe:
Rebased over master and changed the `interpolation_search` logging improvements to avoid needing to pass `(root : string)`.

PR #106, opened 2019-10-08
Fix readonly instances with nonexisting files

pascutto:


src/index.ml
```diff
@@ -236,33 +250,38 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
     in
-    let config = { log_size = log_size * entry_size; readonly } in
+    let config = { log_size = log_size * entry_size; readonly; fresh } in
     let log_path = log_path root in
-    let index_path = index_path root in
-    let log_mem = Tbl.create 1024 in
-    let log = IO.v ~fresh ~readonly ~generation:0L ~fan_size:0L log_path in
-    let generation = IO.get_generation log in
+    let log =
+      if readonly && not (Sys.file_exists log_path) then None
```
CraigFe: Does this lead to weirdness if a read/write instance is opened _after_ a read-only instance has been initialised? It seems that the read-only instance won't see the log.

CraigFe: Hmm, no, scratch that. I see `try_load_log` now :slightly_smiling_face:

test/unix/main.ml
```diff
@@ -286,59 +286,33 @@ module Close = struct
...
+      ]
+    in
+    let check_calls ~readonly =
+      let instance = clone ~readonly in
+      Index.close instance;
+      List.iter
+        (fun (name, call) ->
+          Alcotest.check_raises
+            (Printf.sprintf "%s after close with readonly=%b raises Closed"
+               name readonly) I.Closed (fun () -> call ()))
```
```suggestion
               name readonly) I.Closed call)
```

test/unix/main.ml
```diff
@@ -286,59 +286,33 @@ module Close = struct
...
-    Index.close ro1;
-    test_read_after_close_readonly ro2 k tbl
+  let fail_api_after_close () =
+    let Context.{ clone; _ } = Context.full_index () in
+    let k = Key.v () in
+    let v = Value.v () in
+    let calls t =
+      [
+        ("clear", fun () -> ignore (Index.clear t));
+        ("find", fun () -> ignore (Index.find t k));
```
```suggestion
        ("find", fun () -> ignore (Index.find t k : string));
```
These tests will silently weaken if we were to e.g. change to Lwt or add extra parameters.

A lot of these ignores aren't necessary since they ignore a unit. For `find` and `mem`, I think you should assert the types of the ignored values.

test/unix/main.ml
```diff
@@ -286,59 +286,33 @@ module Close = struct
...
+        ("mem", fun () -> ignore (Index.mem t k));
+        ("replace", fun () -> ignore (Index.replace t k v));
+        ("iter", fun () -> ignore (Index.iter (fun _ _ -> ()) t));
+        ("force_merge", fun () -> ignore (Index.force_merge t));
+        ("flush", fun () -> ignore (Index.flush t));
+      ]
+    in
+    let check_calls ~readonly =
+      let instance = clone ~readonly in
```
```suggestion
      let Context.{ rw; _ } = Context.full_index () in
```

Why clone when you could create a new index altogether?

test/unix/main.ml
icristescu: I think these are already pushed on master [commit](https://github.com/mirage/index/commit/33eca734b570a3fd7036f33196a19d4d8866ac39). 
@pascutto can you do a rebase?

CraigFe: Yep; I was being daft. I've made these suggestions in a separate PR now... #112.

pascutto:
Rebased.

src/index.ml
```diff
@@ -457,34 +490,43 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
-    IO.sync t.log
+    let instance = assert_open t in
+    flush_instance instance
 
   let close it =
     match !it with
     | None -> ()
     | Some t ->
+        it := None;
         t.open_instances <- t.open_instances - 1;
```
pascutto: As discussed with @CraigFe and @icristescu, there is a concurrency issue here. The pattern-matching + set to `None` should be atomic.
The easiest way would be to `Mutex.lock`, that adds a `threads` dependency but it will probably be added anyway with the concurrent merges until we properly functorize over the IO.
Another way is to just merge as-is since thread safety was already killed by https://github.com/mirage/index/pull/108.

samoht: Don't worry about concurrency issues right now -- a comment in the code is enough.

src/index.ml
```diff
@@ -128,11 +131,13 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
   let clear t =
     let t = assert_open t in
```
CraigFe: I know it's not strictly part of this PR, but could change this to `check_open` now? The `assert_*` prefix is now overloaded to mean two different things.

pascutto: Done :heavy_check_mark:

PR #107, opened 2019-10-08
A closed instance now raises an exception when used

pascutto:


src/index.mli
```diff
@@ -68,6 +68,9 @@ exception RO_not_allowed
 (** The exception raised when a write operation is attempted on a read_only
     index. *)
 
+exception Closed
+(** The exception raised when any operation is attempted on a closed index. *)
```
CraigFe: any operation _except_ the `close` operation, which is a no-op on a closed index

CraigFe: (perhaps this behaviour should be tested)

pascutto: Indeed.
I'm not sure how to test it, maybe by re-running the failing operations test on it?

pascutto: Done, thanks :heavy_check_mark:

test/unix/main.ml
```diff
@@ -347,10 +328,8 @@ module Close = struct
       ("replace", `Quick, replace);
       ("open two instances, close one", `Quick, open_readonly_close_rw);
       ("close and reopen on readonly", `Quick, close_reopen_readonly);
-      ("fail to read after close", `Quick, fail_read_after_close);
-      ("fail to write after close", `Quick, fail_write_after_close);
-      ("open twice same instance", `Quick, open_twice);
-      ("open twice same instance readonly", `Quick, open_twice_readonly);
+      ("all operations fail after close", `Quick, fail_api_after_close);
```
```suggestion
      ("non-close operations fail after close", `Quick, fail_api_after_close);
```

PR #108, opened 2019-10-10
use a toplevel buffer for reads

samoht:
spotted by @gs0510

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -150,9 +150,9 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
       let remaining = Int64.sub max offset in
       if remaining <= 0L then ()
       else
-        let size = Stdlib.min remaining page_size in
-        let raw = Bytes.create (Int64.to_int size) in
-        let n = IO.read io ~off:offset raw in
+        let len = Int64.to_int (Stdlib.min remaining page_size) in
+        let raw = Bytes.create len in
+        let n = IO.read io ~off:offset ~len raw in
         let rec read_page page off =
           if off = n then ()
           else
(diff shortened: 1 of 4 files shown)


src/io_array.ml
```diff
@@ -60,12 +60,13 @@ module Make (IO : Io.S) (Elt : ELT) :
 
   let length t = Int64.(div (IO.offset t.io) Elt.encoded_sizeL)
 
+  let buf = Bytes.create 4096
```
CraigFe: Maybe define `max_buffer_size = 4096` to avoid these values becoming desynced

src/index.ml
```diff
@@ -338,8 +338,9 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
   (* Merge [log] with [t] into [dst_io]. [log] must be sorted by key hashes. *)
   let merge_with log index dst_io =
     let entries = 10_000 in
```
gs0510: I know that you did not change this, but why are the number of entries 10,000?

src/unix/index_unix.ml
```diff
@@ -206,9 +206,9 @@ module IO : Index.IO = struct
     t.offset <- t.offset ++ len;
     if t.offset -- t.flushed > auto_flush_limit then sync t
 
-  let read t ~off buf =
+  let read t ~off ~len buf =
```
gs0510: What's the idea behind passing len to read instead of letting read determine the number of bytes it should read?

samoht: We know in advance how many bytes we have to read, so It's to double-check that we have actually read the right amount of data.

CraigFe: It's also no longer true that the number of bytes to be read is equal to the size of the buffer, since the buffer will always be 4096 bytes (and the number of bytes to be read is `<= 4096`).

src/index.ml
pascutto: We want to avoid reading all the entries one by one since this is more costly, so we read them 10_000 by 10_000 so that it is faster.

gs0510: Ahhh I see, why specifically 10,000 though? Just curious :D

pascutto: No specific reason, it seemed okay during the tests/benchs we had.

pascutto:
For the record, we lost the thread-safety with this change.

pascutto:
It's not an *immediate* problem as our current use cases don't require thread safety.
There was also another way which is simply to remove the `pre_fetch` feature, it also improve performances; at least on SSD, not so sure on slow disks.

PR #111, opened 2019-10-14
test: skip file-descriptor test if lsof is unavailable

### LINKED ISSUE #110: Tests are failing when `lsof` is not available
The CI is failing on the debian target because `lsof` is not available on the system.
The code triggering this is [here](https://github.com/mirage/index/blob/master/test/unix/force_merge.ml#L34-L37). The test case should probably be skipped instead of failing.

CraigFe:
Fixes #110.

### DIFF
--- a/test/unix/force_merge.ml
+++ b/test/unix/force_merge.ml
@@ -25,6 +25,7 @@ let test_one_entry r k v =
       Alcotest.failf "Inserted value is not present anymore: %s." k
 
 let test_fd () =
+  let ( >>? ) x f = match x with `Ok x -> f x | err -> err in
   (* TODO: fix these tests to take the correct directory name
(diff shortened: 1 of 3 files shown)


test/unix/force_merge.ml
```diff
@@ -25,49 +25,64 @@ let test_one_entry r k v =
       Alcotest.failf "Inserted value is not present anymore: %s." k
 
 let test_fd () =
+  let ( >>= ) x f = match x with `Ok x -> f x | err -> err in
```
pascutto: Can you pick another infix operator, just to ease the potential migration to `lwt` (if that ever happens).

CraigFe: This function doesn't actually use the Index API at all, but have switched to `>>?` regardless.

PR #113, opened 2019-10-14
Expose closed exception, to be caught by Irmin-pack as well

icristescu:


pascutto:
Right, we should probably do the same with `RO_not_allowed`?

icristescu:
> Right, we should probably do the same with `RO_not_allowed`?

Agree (and done), it makes the exceptions consistent.

CraigFe:
I'm not sure why this is necessary; could you explain it to me?

icristescu:
> I'm not sure why this is necessary; could you explain it to me?

To catch the exception here https://github.com/mirage/irmin/pull/880

CraigFe:
So this is purely to get around the naming conflict between Index and the Index module constructed in Irmin-Pack?

icristescu:
> So this is purely to get around the naming conflict between Index and the Index module constructed in Irmin-Pack?

Well, I guess. Right now Index is only called in Irmin-pack and in the tests, through  
```ocaml 
module Index = Index_unix.Make (Key) (Val)
```
should we rename the module and make Index available ?

pascutto:
I think it makes (more ?) sense to be able to refer to the exceptions using the module that raises them, instead of the one that provides the functor to create the module that raises them.
I am highly unaware of the conventions and best practices about this though.

CraigFe:
I dislike pushing exceptions into functor return types for two reasons:
 - it adds boilerplate: now these exceptions be defined by every functor that constructs an Index (admittedly only one at the moment)

 - it seems to present difficulties w.r.t. exception type equalities, creating the sort of mess that we have with `RO_Not_Allowed` in Irmin_pack where there are multiple redefinitions of the same exception (see https://github.com/mirage/irmin/blob/master/src/irmin-pack/irmin_pack.ml#L87 for instance). I don't actually know for certain that this is part of the problem, however :slightly_smiling_face: 

> should we rename the module and make Index available ?

Renaming the module works, but is invasive and requires changing the entire file to use a less-descriptive name for the Index. My preferred solution is to alias Index (the library) to something in the `Irmin_pack` tests before constructing an `Index` module. e.g.

```
module I = Index
module Index = Index_unix.Make (Key) (Val)
```

Indeed, this is how it's being done the Index tests themselves: https://github.com/mirage/index/blob/master/test/unix/main.ml#L1

CraigFe:
> I think it makes (more ?) sense to be able to refer to the exceptions using the module that raises them, instead of the one that provides the functor to create the module that raises them.
I am highly unaware of the conventions and best practices about this though.

I agree with you that it's semantically nice to only need to think about one module when catching exceptions. 

The corresponding PR in Irmin (https://github.com/mirage/irmin/pull/874/files#diff-1be34172140401295ad7258b1add6782L2133) defines the exception in `irmin.mli`, rather than in the constructed `Store` module. In that case, I think it's definitely the way to go, since there are many, many functors constructing a store.

PR #114, opened 2019-10-14
Use `value Tbl.t` instead of `entry Tbl.t`

pascutto:
We don't need to keep so much information in memory.
- `key` is already the key of the `Tbl.t`, no need to keep it in the value too.
- `key_hash` is only useful during merges, so we can compute it at this time only.
Bonus: now `Tbl.find` has the same return type as `interpolation search`.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -87,13 +87,13 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
 
   exception Invalid_value_size of value
 
-  let append_entry io e =
-    let encoded_key = K.encode e.key in
-    let encoded_value = V.encode e.value in
+  let append_key_value io key value =
+    let encoded_key = K.encode key in
+    let encoded_value = V.encode value in
     if String.length encoded_key <> K.encoded_size then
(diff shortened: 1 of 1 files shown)


src/index.ml
```diff
@@ -454,17 +457,16 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
     if Int64.compare (IO.offset t.log) (Int64.of_int t.config.log_size) > 0
-    then merge ~witness:entry t
+    then merge ~witness:{ key; key_hash = K.hash key; value } t
 
   let iter f t =
     let t = assert_open t in
     Log.debug (fun l -> l "iter %S" t.root);
     if t.config.readonly then sync_log t;
-    Tbl.iter (fun _ e -> f e.key e.value) t.log_mem;
+    Tbl.iter (fun key value -> f key value) t.log_mem;
```
```suggestion
    Tbl.iter f t.log_mem;
```

pascutto: Indeed, done :heavy_check_mark:

PR #115, opened 2019-10-14
Force merge bench

icristescu:
Based on https://github.com/mirage/index/pull/100

### DIFF
--- a/bench/db_bench.ml
+++ b/bench/db_bench.ml
@@ -290,6 +290,35 @@ module Index = struct
     read_amplif 1000;
     [ output ]
 
+  let merge log_nb_entries name =
+    let rw = namespace_v ~fresh:false ~readonly:false name in
+    let t = Index.force_merge rw in
+    Index.await t;
+    Stats.reset_stats ();
+    let rec loop i =
+      if i = log_nb_entries then ()
+      else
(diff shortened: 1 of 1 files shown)


samoht:
Would be very nice to have a flamegraph for what happens during merges!

bench/db_bench.ml
```diff
@@ -468,6 +497,9 @@ let run input json_file_name data_dir metrics_flag =
         [ `OverWrite; `All; `Index ],
         "OverWrite" );
       ((fun () -> lmdb_benchmarks ()), [ `Lmdb; `All ], "Run lmdb benchmarks");
+      ( (fun () -> Index.merge 450_000 "write_random"),
+        [ `Merge; `All; `Index ],
+        "Force merge of \"write_random\" with 450_000/500_000 entries in log" );
```
pascutto: Even more important than the number of entries in the log, it would be useful to know the number of entries in the whole index, since the merge operation depends on the entries in `data` more than the ones in `log`.

icristescu: It is printed at the beginning of each benchmark in `init`, so I think it is redundant to print it here too. I can add it if you want.

pascutto: I see, it's fine then!

PR #116, opened 2019-10-15
Fix merge to prevent duplicates

pascutto:


src/index.mli
```diff
@@ -109,10 +109,13 @@ module type S = sig
...
-  val iter : (key -> value -> unit) -> t -> unit
-  (** Iterates over the index bindings. Order is not specified. In case of
-      recent replacements of existing values (after the last merge), this will
-      hit both the new and old bindings. *)
+  val iter : ?no_duplicates:bool -> (key -> value -> unit) -> t -> unit
+  (** Iterates over the index bindings. Order is not specified.
+      In case of recent replacements of existing values (after the last merge),
+      this will hit both the new and old bindings.
+
+      If [no_duplicates] is [true], then only the most recently added binding
```
CraigFe:
```suggestion
      If [no_duplicates] is [true], only the most recently-added bindings
```

src/index.mli
```diff
@@ -109,10 +109,13 @@ module type S = sig
...
+      will be hit; but this may result in significant cost. *)
```
CraigFe:
```suggestion
      will be hit; this may result in significant cost. *)
```

test/unix/force_merge.ml
```diff
@@ -169,11 +169,35 @@ let readonly_and_merge () =
   loop 10;
   test_fd ()
 
+let test_iter () =
+  let context_with_duplicates () =
+    let { Context.rw; _ } = Context.full_index ~size:4 ~log_size:4 () in
+    let k = Key.v () in
+    let v1 = Value.v () in
+    Index.replace rw k v1;
```
CraigFe: Not strictly a test of `force_merge`, but I think it's OK for it to be in here. Can you add a comment here to say that this `replace` induces a merge operation (and the subsequent one does not)? The interaction between `size` and `log_size` may not be so obvious to a future reader :slightly_smiling_face:

CraigFe:
This PR attempted to do two things:
 - fix the merge logic to prevent duplicate keys from being serialised in the Index;
 - change the API of `iter` to supply a `no_duplicates` flag that could be used to guarantee that the `iter` only saw unique keys. This flag simply caused an initial `force_merge`.

The latter is much harder to achieve if merge operations are able to occur concurrently with `iter` (as enabled by https://github.com/mirage/index/pull/118). In particular, it is no longer possible to guarantee key uniqueness after triggering a merge, because a new merge may occur during the `iter` operation, causing values that have been read in the log to be re-read in the index.

As such, we should delay supporting a `no_duplicates` flag until such a time as its guarantees can be properly specified in the presence of concurrent operations. I have cherry-picked the fix to the merge logic onto the concurrent merges PR (https://github.com/mirage/index/pull/118/commits/520a6a591285403ccbd6e18c35f2af99251b8a8c).

PR #117, opened 2019-10-16
Separate lmdb and index benchmarks

icristescu:
LMDB and index can be benchmarked separately

### DIFF
--- a/bench/db_bench.ml
+++ b/bench/db_bench.ml
@@ -7,6 +7,24 @@ let src = Logs.Src.create "db_bench"
 
 module Log = (val Logs.src_log src : Logs.LOG)
 
+let all = ref false
+
+and with_lmdb = ref false
+
+and with_index = ref false
+
+let speclist =
+  [
+    ("-all", Arg.Set all, "run all benchmarks");
+    ("-lmdb", Arg.Set with_lmdb, "run the lmdb benchmarks");
+    ("-index", Arg.Set with_index, "run the index benchmarks");
+  ]
+
+let usage =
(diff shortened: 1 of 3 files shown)


bench/db_bench.ml
```diff
@@ -278,57 +278,85 @@ module Lmdb = struct
...
          increasing order of keys for lmdb)");
   Index.read_seq rw;
-  Lmdb.read_seq ();
-
   Log.app (fun l -> l "\n");
   Log.app (fun l -> l "Overwrite");
   Index.overwrite rw;
+  Index.close rw
+
+let lmdb_main () =
```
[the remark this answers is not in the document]
pascutto: We should add empty `mli` files everywhere so we can let `dune` warn us about unused functions.

bench/db_bench.ml
```diff
@@ -7,6 +7,24 @@ let src = Logs.Src.create "db_bench"
...
 
+let all = ref false
+
+and with_lmdb = ref false
+
+and with_index = ref false
+
+let speclist =
+  [
+    ("-all", Arg.Set all, "run all benchmarks");
```
samoht: Can you use `--` instead of `-`?

samoht: But this could be done in a separate PR using cmdliner

icristescu: I hesitated between Arg and Cmdliner, why do we prefer the latter?

samoht: because it's generating a nice `--help` document too :-) (and we are using it in all the other MirageOS projects)

PR #118, opened 2019-10-17
Concurrent merges

### LINKED ISSUE #13: Have non-blocking merges

pascutto:
Fixes #13.
On top of #105 to ease debugging.
This is still very WIP.

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -2,4 +2,4 @@
  (public_name index)
  (name index)
  (modules_without_implementation io)
- (libraries logs fmt))
+ (libraries logs logs.threaded fmt threads))
--- a/src/dune
+++ b/src/dune
@@ -2,4 +2,4 @@
  (public_name index)
  (name index)
  (modules_without_implementation io)
- (libraries logs logs.threaded fmt threads))
+ (libraries logs fmt))
--- a/src/index.ml
+++ b/src/index.ml
@@ -406,7 +406,8 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
         else no_changes ()
 
   let find_instance t key =
-    let find_if_exists ~name ~find = function
+    let find_if_exists ~name ~find db () =
+      match db with
       | None ->
           Log.debug (fun l ->
               l "[%s] %s is not present" (Filename.basename t.root) name);
@@ -417,39 +418,26 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
               l "[%s] found in %s" (Filename.basename t.root) name);
           ans
     in
+    let ( @~ ) a b = try a () with Not_found -> b () in
     let find_log_index () =
-      try find_if_exists ~name:"log" ~find:(fun log -> Tbl.find log.mem) t.log
-      with Not_found -> (
-        match
-          find_if_exists ~name:"index" ~find:interpolation_search t.index
-        with
-        | Some e -> e
-        | None -> raise Not_found )
+      find_if_exists ~name:"log" ~find:(fun log -> Tbl.find log.mem) t.log
+      @~ find_if_exists ~name:"index" ~find:interpolation_search t.index
     in
     IO.Mutex.with_lock t.rename_lock (fun () ->
         if t.config.readonly then sync_log t;
-        try
-          find_if_exists ~name:"log" ~find:(fun log -> Tbl.find log.mem) t.log
-        with Not_found -> (
-          try
-            find_if_exists ~name:"log_async"
-              ~find:(fun log -> Tbl.find log.mem)
-              t.log_async
-          with Not_found -> (
-            match
-              find_if_exists ~name:"index" ~find:interpolation_search t.index
-            with
-            | Some e -> e
-            | None when t.config.readonly ->
-                sync_log t;
-                find_log_index ()
-            | None -> raise Not_found ) ))
+        find_if_exists ~name:"log_async"
+          ~find:(fun log -> Tbl.find log.mem)
+          t.log_async
+        @~ fun () ->
+        find_log_index @~ fun () ->
+        if t.config.readonly then (
+          sync_log t;
+          find_log_index () )
+        else raise Not_found)
 
   let find t key =
     let t = check_open t in
-    Log.info (fun l ->
-        l "[%s] ro= %b find %a" (Filename.basename t.root) t.config.readonly
-          K.pp key);
+    Log.info (fun l -> l "[%s] find %a" (Filename.basename t.root) K.pp key);
     find_instance t key
 
   let mem t key =
--- a/src/index.ml
+++ b/src/index.ml
@@ -356,7 +356,7 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
       Some { io; mem } )
     else None
 
-  let sync_log ?(force = false) t =
+  let sync_log t =
     Log.debug (fun l ->
         l "[%s] checking for changes on disk" (Filename.basename t.root));
     let no_changes () =
@@ -403,22 +403,6 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
           (* In that case the log has probably been emptied and is being
              refilled with async_log contents. *)
           no_changes ()
-        else if force then (
-          Log.debug (fun l -> l "[%s] force sync" (Filename.basename t.root));
-          Tbl.clear log.mem;
-          iter_io add_log_entry log.io;
-          may (fun (i : index) -> IO.close i.io) t.index;
-          let index_path = index_path t.root in
-          if Sys.file_exists index_path then
-            let io =
-              IO.v ~fresh:false ~readonly:true ~generation ~fan_size:0L
-                index_path
-            in
-            let fan_out =
-              Fan.import ~hash_size:K.hash_size (IO.get_fanout io)
-            in
-            t.index <- Some { fan_out; io }
-          else () )
         else no_changes ()
 
   let find_instance t key =
@@ -457,7 +441,7 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
             with
             | Some e -> e
             | None when t.config.readonly ->
-                sync_log ~force:true t;
+                sync_log t;
                 find_log_index ()
             | None -> raise Not_found ) ))
 
--- a/src/index.ml
+++ b/src/index.ml
@@ -122,8 +122,8 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
     mutable log_async : log option;
     mutable open_instances : int;
     lock : IO.lock option;
-    mutable merge_lock : Mutex.t;
-    mutable rename_lock : Mutex.t;
+    mutable merge_lock : IO.Mutex.t;
+    mutable rename_lock : IO.Mutex.t;
   }
 
   type t = instance option ref
@@ -136,23 +136,22 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
     Log.debug (fun l -> l "clear %S" t.root);
     if t.config.readonly then raise RO_not_allowed;
     t.generation <- 0L;
-    Mutex.lock t.merge_lock;
-    let log = assert_and_get t.log in
-    IO.clear log.io;
-    Tbl.clear log.mem;
-    may
-      (fun l ->
-        IO.clear l.io;
-        IO.close l.io)
-      t.log_async;
-    may
-      (fun (i : index) ->
-        IO.clear i.io;
-        IO.close i.io)
-      t.index;
-    t.index <- None;
-    t.log_async <- None;
-    Mutex.unlock t.merge_lock
+    IO.Mutex.with_lock t.merge_lock (fun () ->
+        let log = assert_and_get t.log in
+        IO.clear log.io;
+        Tbl.clear log.mem;
+        may
+          (fun l ->
+            IO.clear l.io;
+            IO.close l.io)
+          t.log_async;
+        may
+          (fun (i : index) ->
+            IO.clear i.io;
+            IO.close i.io)
+          t.index;
+        t.index <- None;
+        t.log_async <- None)
 
   let flush_instance instance =
     Log.debug (fun l ->
@@ -164,9 +163,7 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
   let flush t =
     let instance = check_open t in
     Log.info (fun l -> l "[%s] flush" (Filename.basename instance.root));
-    Mutex.lock instance.rename_lock;
-    flush_instance instance;
-    Mutex.unlock instance.rename_lock
+    IO.Mutex.with_lock instance.rename_lock (fun () -> flush_instance instance)
 
   let ( // ) = Filename.concat
 
@@ -330,8 +327,8 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
       root;
       index;
       open_instances = 1;
-      merge_lock = Mutex.create ();
-      rename_lock = Mutex.create ();
+      merge_lock = IO.Mutex.create ();
+      rename_lock = IO.Mutex.create ();
       lock;
     }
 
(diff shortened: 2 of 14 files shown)


src/index.ml
```diff
@@ -450,56 +496,82 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+      Fan.finalize index.fan_out;
+      IO.set_fanout merge (Fan.export index.fan_out);
+      Mutex.lock t.rename_lock;
+      IO.rename ~src:merge ~dst:index.io;
+      t.index <- Some index;
+      let log_async = assert_and_get t.log_async in
+      IO.clear ~keep_generation:true log.io;
+      Tbl.iter (append_key_value log.io) log_async.mem;
+      IO.sync log.io;
+      IO.set_generation log.io generation;
```
samoht: why setting the generation so late? would be better to do this just after the clear I think

src/unix/index_unix.ml
```diff
@@ -193,6 +193,7 @@ module IO : Index.IO = struct
     sync src;
     Unix.close dst.raw.fd;
     Unix.rename src.file dst.file;
+    Buffer.reset dst.buf;
```
samoht: this might be allocating a lot.

src/index.ml
```diff
@@ -143,12 +146,28 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+    Log.debug (fun l ->
+        l "[%s] flushing instance" (Filename.basename instance.root));
+    if instance.config.readonly then raise RO_not_allowed;
+    let log = assert_and_get instance.log in
+    IO.sync log.io
+
+  let flush t =
+    let instance = check_open t in
+    Log.info (fun l -> l "[%s] flush" (Filename.basename instance.root));
+    Mutex.lock instance.merge_lock;
```
samoht: why is this blocked during a merge?

src/dune
```diff
@@ -2,4 +2,4 @@
  (public_name index)
  (name index)
  (modules_without_implementation io)
- (libraries logs fmt))
+ (libraries logs logs.threaded fmt threads))
```
samoht: is there any hope to expose the right hooks in `IO` and not depend directly on threads in the core library?

src/index.ml
```diff
@@ -409,56 +512,86 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
-      |> ignore;
-      Array.fast_sort compare_entry b;
-      b
-    in
-    let fan_size =
-      match t.index with
-      | None -> Tbl.length log.mem
-      | Some index ->
-          (Int64.to_int (IO.offset index.io) / entry_size) + Tbl.length log.mem
+    flush_instance t;
```
icristescu: Maybe you want to call `flush`(with the `rename_lock`) here and not `flush_instance`?

src/index.ml
```diff
@@ -409,56 +512,86 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+      t.generation <- generation;
+      Tbl.iter
+        (fun key value ->
+          Tbl.replace log.mem key value;
+          append_key_value log.io key value)
+        log_async.mem;
+      IO.sync log.io;
+      t.log_async <- None;
+      Mutex.unlock t.rename_lock;
+      IO.clear log_async.io;
```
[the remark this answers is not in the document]
pascutto: `t.log_async` is an option containing `log_async` (before line 588).
The reason why I set `t.log_async` to `None` before `clear` and `close` is I would like to maintain the invariant that if `t.async` is `Some log_async`, then `log_async` is still usable. Why do you think it's better to do it the other way?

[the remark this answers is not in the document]
icristescu: It's a detail, but I don't see why you don't close and clear inside the lock? you don't need the extra variable then and there is no clean up to do outside the lock.

src/index.ml
```diff
@@ -409,56 +512,86 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+            merge_with log_array index merge;
+            index
+      in
+      Fan.finalize index.fan_out;
+      IO.set_fanout merge (Fan.export index.fan_out);
+      Mutex.lock t.rename_lock;
+      IO.rename ~src:merge ~dst:index.io;
+      t.index <- Some index;
+      let log_async = assert_and_get t.log_async in
+      IO.clear ~keep_generation:true log.io;
```
icristescu: Why keep the generation in `log.io` here, since you change it two lines below? 
I think that you don't need to have the generation in `log_async`, no?

src/index.ml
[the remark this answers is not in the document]
pascutto: Sure, I wanted to keep the locked portion as small as possible, and I don't think this part is critical, but if that helps in any way I can do that inside the lock.

src/index.ml
pascutto: We have to keep in mind that a readonly instance might read this at any time. If a readonly instance reads the index while the generation is set to zero (which happens if we don't keep the generation because it will be reset), it will believe that our index is at generation 0 and in particular won't read the `index` (but perhaps we can change this behaviour if that doesn't seem right).

pascutto: And about where the generation is needed, strictly speaking we are only reading the generation in the `log` file, but having this header meaningful everywhere can help debugging, because it allows us to understand at what stage of a merge (or not) we are. That can be useful when writing a diagnosis tool.
I don't think this part of the code involves the generation in `log_async` however.

samoht:
What about adding hooks to ease testing -- maybe a lock that the testing framework can hold to block the merge while doing other operations, to check that there are progress going on.

More generally, I think we should try to redesign the library a bit to make it more easily testable -- I know @CraigFe started to work a bit on this on the I/O side, but would be great if we could push this "design-to-ease-tests" at the core of the library, especially for the merges.

icristescu:
Update on debugging: 
- from the logs, it seems that the `readonly_and_merge` tests fail when a write occurs at the end of a merge. In this case it is expected to fail, a flush is needed between a write and a read of a RO instance. I've updated the tests, and cannot reproduce an error for this particular tests. A test to reproduce this failing behaviour is `_write_after_merge`.
- tests `readonly_s` and `readonly` still fail and it seems (again from reading the logs) that both fail on the same case: `log_async` is written during a merge and moved into `log` at the end of the merge. But between the copy of `log_async` to `log` and the `Io.sync log` the log becomes empty.

icristescu:
> * tests `readonly_s` and `readonly` still fail and it seems (again from reading the logs) that both fail on the same case: `log_async` is written during a merge and moved into `log` at the end of the merge. But between the copy of `log_async` to `log` and the `Io.sync log` the log becomes empty. The merge is concurrent with a `find` of a RO instance.

This is now fixed by the last commit. The issue was a buffer shared between an RO and an RW instance. But there is (at least) one bug remaining.

icristescu:
> But there is (at least) one bug remaining.

This bug was caused by something like* the following scenario:
RW - writes value in log, then calls `merge`, the value is moved into index, but the generation is not set
RO - calls `sync_log` and reads generation 
RW - clears log, sets generation 
RO - reads offset of log but does not see any changes in neither generation nor offset. it does not update its index and therefore cannot find the value. 

I proposed a quick fix which is to force an update of the index when the value is not found. This is only to check that the bug is fixed, maybe we can improve it. 

*something like this, because is hard to track what really happens as too many logs introduce delays and the bug does not reappear.

CraigFe:
After the most-recent commit (https://github.com/mirage/index/pull/118/commits/e76715cb7352b41948d1f8d3527801b638b926b3), I have gone through 1000 consecutive test runs without any failures; this seems to have fixed something at least :tada:

src/index.ml
```diff
@@ -364,42 +399,76 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+              ~find:(fun log -> Tbl.find log.mem)
+              t.log_async
+          with Not_found -> (
+            match
+              find_if_exists ~name:"index" ~find:interpolation_search t.index
+            with
+            | Some e -> e
+            | None when t.config.readonly ->
+                sync_log ~force:true t;
+                find_log_index ()
```
pascutto: So in that case, if a read-only instance does not find the binding, it reads the whole `log` and re-imports the fanout again from a new index file descriptor, then retries to read, do I understand correctly?
Why is this necessary?

icristescu: The idea is that if the last steps of a merge interleave with a RO's sync_log, the RO instance does not see the latest update. 
 
> This bug was caused by something like* the following scenario:
- start with an empty log, for both RW and RO
> RW - writes value in log, then calls `merge`, the value is moved into index, but the generation is not set
> RO - calls `sync_log` and reads generation 

-`IO.get_generation`
> RW - clears log, sets generation
> RO - reads offset of log 

-`IO.force_offset`
> RO does not see any changes in neither generation nor offset. it does not update its index and therefore cannot find the value.

icristescu: But this situation can also appear if you start with a log of size 1 and add one entry in log_async during the merge. By the end of the merge you end up with a log of size 1, and RO could miss that there was an update.

src/index.ml
```diff
@@ -417,39 +418,27 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
-                sync_log t;
-                find_log_index ()
-            | None -> raise Not_found ) ))
+        find_if_exists ~name:"log_async"
+          ~find:(fun log -> Tbl.find log.mem)
+          t.log_async
+        @~ fun () ->
+        find_log_index
+        @~
+        if t.config.readonly then (
```
CraigFe: Doesn't this whole `if` need to be inside the `fun () ->`, else you're doing `sync_log` even if `find_log_index` returns the value?

src/io_array.ml
```diff
@@ -78,7 +78,7 @@ module Make (IO : Io.S) (Elt : ELT) :
       Log.warn (fun m ->
           m "Requested pre-fetch region is empty: [%Ld, %Ld]" low high)
     else if range > max_buffer_size then
-      Log.debug (fun m ->
+      Log.warn (fun m ->
```
CraigFe: Why is this a `warn`? Doesn't this happen frequently since `interpolation_search` optimistically pre-fetches?

src/io_array.ml
```diff
@@ -89,21 +89,13 @@ module Make (IO : Io.S) (Elt : ELT) :
               ( div b.low_off Elt.encoded_sizeL,
                 div b.high_off Elt.encoded_sizeL )
           in
-          if low >= low_buf && high <= high_buf then
-            Log.debug (fun m ->
```
CraigFe: Personally, I found these debug messages helpful for tracking down issues in the `interpolation_search`. I assume you removed them because they were too noisy?

src/search.ml
```diff
@@ -83,39 +83,37 @@ module Make
...
-        if not Metric.(key_metric = e_metric) then None
-        else if Key.equal (Entry.to_key e) key then Some (Entry.to_value e)
+        if not Metric.(key_metric = e_metric) then raise Not_found
+        else if Key.equal (Entry.to_key e) key then Entry.to_value e
         else (search [@tailcall]) op i
     in
-    match search Int64.succ index with
-    | Some e -> Some e
-    | None -> (search [@tailcall]) Int64.pred index
+    (search [@tailcall]) Int64.pred index
```
CraigFe: This now only searches backwards: have we changed something that makes searching forwards unnecessary? If so, I think it's probably worth renaming `look_around` to reflect that it's only looking in one direction.

src/unix/index_unix.ml
```diff
@@ -175,9 +175,6 @@ module IO : Index.IO = struct
     else (
       Raw.unsafe_write t.raw ~off:t.flushed buf;
       Raw.Offset.set t.raw offset;
-
-      (* concurrent append might happen so here t.offset might differ
```
CraigFe: It's no longer clear to me from reading the code why this is a failure case.

src/io_array.ml
pascutto: I think the fanout is supposed to create < 4096 bytes buckets, so the range of search of the interpolation search should always be smaller.

src/unix/index_unix.ml
pascutto: Actually, concurrent append cannot happen anymore due to the lock file, the comment was outdated.
The check still stands: after writing the file, if the previously flushed data + the data we just flushed is not the same size as the content written to the disk (header + data), then something is wrong.

CraigFe: I see; I didn't understand the purpose of the check then. What sort of runtime behaviour might cause this check to fail? Should it be an assert instead?

pascutto: I can't think of a scenario where this fails (but maybe there is?), and it could indeed be an assert. If I remember well, we had issues with this when the lock file wasn't there yet, and we needed to print for debugging.

src/search.ml
pascutto: Oops my bad, good catch!

CraigFe: As an aside, we should have tests that expose this behaviour using a degenerate hash function.

src/index.ml
```diff
@@ -332,14 +368,38 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+            let io =
+              IO.v ~fresh:false ~readonly:true ~generation:0L ~fan_size:0L path
+            in
+            t.log_async <- Some { log with io };
+            let new_log_offset = IO.offset io in
+            if log_offset <> new_log_offset then (
+              Tbl.clear log.mem;
+              iter_io (add_log_entry log) io ) )
+          else ()
+        with IO.Bad_Fd_Read ->
```
CraigFe: Perhaps this exception should not have `Fd` in its name: that's a detail of the  `Index_unix` usecase. It also hasn't been documented / explained anywhere (e.g. in `src/io.mli`).

src/index.ml
```diff
@@ -332,14 +368,38 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+          ()
+          (* if log_async does not exits anymore, then its contents have been moved to log and the generation has changed *)
```
```suggestion
          (* if log_async does not exist anymore, then its contents have been moved to log and the generation has changed *)
```
Probably want an 80-char word wrap here too.

src/index.ml
```diff
@@ -364,42 +424,51 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
               l "[%s] new entries detected, reading log from disk"
                 (Filename.basename t.root));
           iter_io add_log_entry log.io ~min:log_offset )
-        else if log_offset > new_log_offset then assert false
+        else if log_offset > new_log_offset then
+          (* In that case the log has probably been emptied and is being
+             refilled with async_log contents. *)
+          no_changes ()
```
CraigFe: Seems odd to have two `else` cases that both do `no_changes ()`: is this just so that the comment can be here?

src/index.ml
```diff
@@ -560,34 +668,30 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
+            | Some log -> Tbl.iter f log.mem );
+            may
+              (fun (i : index) -> iter_io (fun e -> f e.key e.value) i.io)
+              t.index)
 
   let close it =
     match !it with
     | None -> Log.info (fun l -> l "close: instance already closed")
     | Some t ->
         (* XXX This piece of code is not thread safe. *)
```
CraigFe: Perhaps this comment no longer belongs here.

src/unix/index_unix.ml
```diff
@@ -366,6 +369,24 @@ module IO : Index.IO = struct
...
+      try
+        let ans = f () in
+        Mutex.unlock t;
+        ans
+      with e ->
+        Mutex.unlock t;
+        raise e
+  end
+
+  let () = Logs_threaded.enable ()
```
CraigFe: This seems odd; shouldn't the logs be enabled elsewhere?

src/index.ml
pascutto: Yes I think it's worth having a specific case here to highlight some action might be needed (or a comment saying nothing is needded).

pascutto:
Fixed :heavy_check_mark:

PR #119, opened 2019-10-21
Timer merge

icristescu:
Rebased over PR #100.
 
With the results:
```
Fill in random order
index: 2.824067 micros/op; 	 354099.247645 op/s; 	 15.934466 MB/s; 	 total time = 28.240670s.
	write amplification in bytes = 10.505701; in nb of writes = 0.000587;
	merge time = 23.098916; percentage = 0.817931;
```

### DIFF
--- a/bench/db_bench.ml
+++ b/bench/db_bench.ml
@@ -91,7 +91,8 @@ let print_results db f =
   let ops_sec = float_of_int nb_entries /. time in
   Log.app (fun l ->
       l "%s: %f micros/op; \t %f op/s; \t %f MB/s; \t total time = %fs." db
-        sec_op ops_sec mb time)
+        sec_op ops_sec mb time);
+  time
 
 module Index = struct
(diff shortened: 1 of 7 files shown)


pascutto:
If you want to call it `percentage` it's probably better to have it multiplied by 100 before.

CraigFe:
Metrics already has a timing mechanism in the form of the `run` function, which seems more flexible in the sense that it allows the reporter to determine the time operation used. Is there a particular reason for not using that here?

icristescu:
> Metrics already has a timing mechanism in the form of the `run` function, which seems more flexible in the sense that it allows the reporter to determine the time operation used. Is there a particular reason for not using that here?

We removed metrics from the main code base, and moved it to the benchs only, so we can't call metrics from the merge function.

icristescu:
Not pertinent anymore, as merges are now done in separate threads.

PR #120, opened 2019-10-22
IO: protect write attempt in readonly mode

phink:
In the IO module, it looks like when a new file is created in `readonly` mode ([here](https://github.com/mirage/index/blob/master/src/unix/index_unix.ml#L278)), some [write attempts are performed](https://github.com/mirage/index/blob/master/src/unix/index_unix.ml#L299).
Here is a simple piece of code that reproduces this behaviour.

```ocaml
let _ =
  Index_unix.Private.IO.v
    ~readonly:true
    ~fresh:true
    ~generation:0L
    ~fan_size:0L
    "some_file_that_does_not_exist"
```

This PR aims to avoid the file to be created and the following bad descriptor unix error `Unix.Unix_error(Unix.EBADF, "read", "")` to be thrown by mimicking the failure raised in the [second branch](https://github.com/mirage/index/blob/master/src/unix/index_unix.ml#L308) before calling `Unix.openfile`.

### DIFF
--- a/src/unix/index_unix.ml
+++ b/src/unix/index_unix.ml
@@ -294,13 +294,15 @@ module IO : Index.IO = struct
     mkdir (Filename.dirname file);
     match Sys.file_exists file with
     | false ->
-        let x = Unix.openfile file Unix.[ O_CREAT; O_CLOEXEC; mode ] 0o644 in
-        let raw = Raw.v x in
-        Raw.Offset.set raw 0L;
-        Raw.Fan.set_size raw fan_size;
-        Raw.Version.set raw;
-        Raw.Generation.set raw generation;
-        v ~fan_size ~offset:0L ~version:current_version raw
+        if readonly then Fmt.failwith "IO.v: cannot set a readonly file"
+        else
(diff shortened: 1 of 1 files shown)


pascutto:
Right, the check is actually done elsewhere (namely [here](https://github.com/mirage/index/blob/master/src/index.ml#L266)). The IO module is not intended to be used apart in the tests or benchs (that's why it is under the `Private` module.
Did you experience this issue while using the higher-level API?

phink:
I experienced the issue with the current version of `tezos` and I thought we were almost up-to-date with index, but I did not see your last work on `readonly` files. It might be time to pick some commits! 
Sorry for the noise, you can close the PR if you think performing the check in IO is useless.

pascutto:
Indeed, the handling of read-only instances is slightly better since #106 so you shouldn't experience this issue anymore (let us know if you do).
Don't close the PR, I think it makes sense to have `IO` consistent and there is an `IO` refactoring planned soon, so it might be useful.

pascutto:
The IO refactor will try to focus on type-encoded capabilities, so we'll take special care about this concern. I'm closing this PR in the meantime.

PR #122, opened 2019-11-06
Add benchmarks for absent values

icristescu:


bench/db_bench.ml
```diff
@@ -74,6 +74,22 @@ let print_results db f =
       l "%s: %f micros/op; \t %f op/s; \t %f MB/s; \t total time = %fs." db
         sec_op ops_sec mb time)
 
+let rec random_new_key ar =
+  let k = Context.Key.v () in
+  if Array.exists (fun (k', _) -> k = k') ar then random_new_key ar else k
+
+let populate_absents ar nb_entries =
+  let absents = Array.make nb_entries ("", "") in
```
pascutto: How about using a `Hashtbl.t` to maje it more efficient ?

icristescu: We are not timing this, so it's ok if it's slow. 
I switch from hash tables to arrays, because I think arrays have less allocation. Memory allocations make a difference when benchmarking: the more memory is occupied the slower the operations we want to benchmark.

pascutto:
Is this ready to merge ?

icristescu:
> Is this ready to merge ?

yes

PR #128, opened 2019-11-14
Call `fsync` on `flush`

### LINKED ISSUE #126: Add an option to force fsync on flush

pascutto:
Fixes #126.
I'm not sure why this should be an option though, I feel like calling `flush` we expect the content to be... flushed. Exposing some details of our internal caches doesn't feel right to me, and I can't think of a scenario where the use calls `flush` but agrees that some data is still not written to the disk, so this PR calls `fsync` on every `flush`.
Let me know if I got something wrong about it, maybe @samoht you have some insights about it?

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -72,7 +72,7 @@ module type S = sig
 
   val iter : (key -> value -> unit) -> t -> unit
 
-  val flush : t -> unit
+  val flush : ?with_fsync:bool -> t -> unit
 
   val close : t -> unit
 end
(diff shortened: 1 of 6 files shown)


samoht:
There are various consistency and durability models :-) see lmdb and/or the doc for open. Sometimes it’s fine to just be consistent but not durable, so we should let users have that option

edwintorok:
It would be useful to do an `fsync` in rename too.
See https://thunk.org/tytso/blog/2009/03/15/dont-fear-the-fsync/ and https://www.kernel.org/doc/Documentation/filesystems/ext4.txt. Ext4 has some code to detect "broken applications" that do not use fsync before renaming, but not all FS may do:
> Many broken applications don't use fsync() when 
		replacing existing files via patterns such as
			fd = open("foo.new")/write(fd,..)/close(fd)/
			rename("foo.new", "foo"), or worse yet,
			fd = open("foo", O_TRUNC)/write(fd,..)/close(fd).
			If auto_da_alloc is enabled, ext4 will detect
			the replace-via-rename and replace-via-truncate
			patterns and force that any delayed allocation
			blocks a

For just flushing the buffers `fdatasync` may be a lower overhead option (it writes out only the data, and not file metadata like updated `mtime`)

pascutto:
Rebased, added `fsync` as an option only, and applied it before renaming, as well as on `merge` and `close` since those are not frequent and performing a fsync for those would require the user to manually specify the files, since you can't `flush` after `close`, and `flush` does not flush the `data` file..
Comments welcome on naming and documentation.

PR #130, opened 2019-11-18
Breaking up the big bench.

gs0510:
Probably very terrible ocaml, and not the best way to split up benchmarks, but it's a start. Eventually we will have regex matching, but for now this works! 

(I spent two days trying to get regex to work, gave up and went the easier route :))

### DIFF
--- a/bench/main.ml
+++ b/bench/main.ml
@@ -1,4 +1,5 @@
 open Common
+open Cmdliner
 
 let key_size = 32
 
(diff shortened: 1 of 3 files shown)


samoht:
Looks great, thanks!

bench/main.ml
```diff
@@ -43,17 +44,47 @@ let rec finds t count = function
...
+  | "read_write" -> read_write_bench ()
+  | "all" ->
+      write_bench ();
+      read_bench ();
+      read_write_bench ()
+  | _ -> raise (UnsupportedCommand "input not supported")
+
+let input =
+  let doc = "options: write, read, read_write; all" in
+  Arg.(value & opt string "" & info [ "b"; "bench" ] ~doc)
```
samoht: actually this probably needs a better default (e.g. `"all"`). I'll push a fix.

PR #134, opened 2019-11-27
Split db benchmarks

gs0510:
There's a lot of benchmarks for db!
This PR is mostly trying to separate them so they can be run individually and also moving out the lmdb benchmarks to their own file.

It is ~not~ pretty, and regex would a LOT better solution, but it is also hard :)

### DIFF
--- a/bench/db_bench.ml
+++ b/bench/db_bench.ml
@@ -7,24 +7,6 @@ let src = Logs.Src.create "db_bench"
 
 module Log = (val Logs.src_log src : Logs.LOG)
 
-let all = ref false
-
-and with_lmdb = ref false
-
-and with_index = ref false
-
-let speclist =
-  [
-    ("-all", Arg.Set all, "run all benchmarks");
-    ("-lmdb", Arg.Set with_lmdb, "run the lmdb benchmarks");
-    ("-index", Arg.Set with_index, "run the index benchmarks");
-  ]
-
-let usage =
-  "usage: " ^ Sys.argv.(0)
-  ^ " [-all | -lmdb | -index], if no argument provided then minimal index \
-     benchmarks are run"
-
 let seed = 1
 
 let () = Random.init seed
@@ -315,6 +297,22 @@ module Lmdb = struct
   let close env = closedir env
 end
 
+let lmdb_benchmarks () =
+  Log.app (fun l -> l "\n Fill in increasing order of keys");
+  Lmdb.fail_on_error Lmdb.write_seq;
+  populate ();
+  Log.app (fun l -> l "\n Fill in random order and sync after each write");
+  Lmdb.fail_on_error Lmdb.write_sync;
+  Log.app (fun l -> l "\n Fill in random order");
+  let lmdb, env = R.get_ok (Lmdb.write_random ()) in
+  Log.app (fun l -> l "\n Read in random order ");
+  Lmdb.read_random lmdb;
+  Log.app (fun l -> l "\n Read in increasing order of keys");
+  Lmdb.read_seq ();
+  Log.app (fun l -> l "\n Overwrite");
+  Lmdb.overwrite lmdb;
+  Lmdb.close env
+
 let init () =
   Common.report ();
   Index.init ();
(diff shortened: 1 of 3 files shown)


gs0510:
@CraigFe 
@icristescu

bench/lmdb_bench.mli
```diff
@@ -0,0 +1 @@
+(** Left empty on purpose *)
```
CraigFe:
```suggestion
(* left empty on purpose *)
```
No reason for this to be a doc-comment.

bench/db_bench.ml
```diff
@@ -219,188 +198,127 @@ module Index = struct
...
-  Index.read_seq rw;
-  Log.app (fun l -> l "\n");
-  Log.app (fun l -> l "Overwrite");
-  Index.overwrite rw;
+  let () =
+    match input with
+    | `Find `RW | `IndexAll | `All | `Minimal ->
+        let () = Log.app (fun l -> l "\n RW Read in random order") in
+        Index.read_random rw
+    | _ -> ()
```
CraigFe: How about extracting this logic to a function? It's used in quite a lot of places. Maybe something like:
```ocaml
val may_run : 
    ~bench:(unit -> unit) ->
    ~name:string ->
    ~triggers:bench_id list ->
    ~input:bench_id -> 
    unit
```
Then maybe you could store all of the benchmarks in a list and `List.iter` over it, rather than repeatedly matching on the input.

bench/db_bench.ml
```diff
@@ -116,8 +95,8 @@ module Index = struct
     in
     let ratio_reads = float_of_int stats.nb_reads /. float_of_int nb_entries in
     Log.app (fun l ->
-        l "\tread amplification in bytes = %f; in nb of reads = %f "
-          ratio_bytes ratio_reads)
+        l "\tread amplification in bytes = %f; in nb of reads = %f " ratio_bytes
+          ratio_reads)
```
CraigFe: Are you using the right version of OCamlformat?

bench/dune
```diff
@@ -6,7 +6,12 @@
 (executable
  (name db_bench)
  (modules db_bench)
- (libraries fmt index.unix common lmdb bigstring rresult logs logs.fmt))
+ (libraries fmt index.unix common bigstring logs logs.fmt cmdliner))
+
+(executable
```
CraigFe: It's not clear to me why you would want these to be two separate binaries; did you have a particular reason for doing it this way?

bench/db_bench.ml
```diff
@@ -219,188 +198,127 @@ module Index = struct
...
-  Log.app (fun l -> l "Fill in increasing order of hashes");
-  Index.write_seq_hash ();
-  Log.app (fun l -> l "\n");
-  Log.app (fun l -> l "Fill in decreasing order of hashes");
-  Index.write_rev_seq_hash ();
-  populate ();
-  Log.app (fun l -> l "\n");
-  Log.app (fun l -> l "Fill in random order and sync after each write");
-  Index.write_sync ();
+open Cmdliner
```
CraigFe: You can move this `open` to below the `run` function to encourage a clean separation between the runtime API and the Cmdliner API.

bench/db_bench.ml
gs0510: Yep! 0.12

bench/dune
gs0510: I figured since I split the two benchmarks, might as well split the binary too. I don't have particularly strong opinions on this.

bench/lmdb_bench.ml
```diff
@@ -0,0 +1,237 @@
...
+  Log.app (fun l -> l "Entries: %d." nb_entries);
+  Log.app (fun l -> l "Log size: %d." log_size);
+  populate ()
+
+let run input =
+  init ();
+  Log.app (fun l -> l "\n Fill in random order");
+  let lmdb, env = R.get_ok (Lmdb.write_random ()) in
+  let match_input ~bench ~triggers ~message =
+    List.iter
```
CraigFe: Might want `if List.mem input triggers` instead.

bench/lmdb_bench.ml
```diff
@@ -0,0 +1,237 @@
...
+      (fun trigger ->
+        if input = trigger then
+          let () = Log.app (fun l -> l "\n %s" message) in
+          bench ()
+        else ())
+      triggers
+  in
+  let () =
+    match_input
```
CraigFe: Nice :slightly_smiling_face: Now you can keep `(unit -> unit, trigger list, string)` tuples in a list and do something like:
```ocaml
List.iter (fun (bench, triggers, message) -> match_input ~bench ~triggers ~message)
```
Keeping specifications in data rather than code etc. etc.

bench/lmdb_bench.ml
```diff
@@ -0,0 +1,237 @@
...
+  let _, time = with_timer f in
+  let micros = time *. 1_000_000. in
+  let sec_op = micros /. float_of_int nb_entries in
+  let mb = float_of_int (entry_size * nb_entries / 1_000_000) /. time in
+  let ops_sec = float_of_int nb_entries /. time in
+  Log.app (fun l ->
+      l "%s: %f micros/op; \t %f op/s; \t %f MB/s; \t total time = %fs." db
+        sec_op ops_sec mb time)
+
+module Lmdb = struct
```
icristescu: There are a few things in the `lmdb_bench.ml` that are common to the `index_bench.ml`: can you keep only the `lmdb` module here, and then call it from the `index_bench`?  Then you will also have only one executable. 

In particular the data set I think should be shared between the different benchmarks.

bench/db_bench.ml
```diff
@@ -327,80 +325,87 @@ let init () =
...
+      ( (fun () -> Index.ro_read_random rw),
+        [ `Read `Ro; `All; `Minimal; `Index ],
+        "RO Read in random order" );
+      ( (fun () -> Index.read_absent rw),
+        [ `Read `Absent; `All; `Minimal; `Index ],
+        "Read 1000 absent values" );
+      ( (fun () -> Index.read_seq rw),
+        [ `Read `Seq; `All; `Index ],
+        "Read in sequential order (increasing order of hashes for index" );
+      ( (fun () -> Index.write_seq ()),
```
```suggestion
      (Index.write_seq),
```
and similarly below

PR #135, opened 2019-11-27
Clear log mem on close

icristescu:
Similarly to `close` in Irmin-pack.

### DIFF
--- a/test/unix/main.ml
+++ b/test/unix/main.ml
@@ -139,28 +139,28 @@ end
 module DuplicateInstance = struct
   let find_present () =
     let Context.{ rw; tbl; clone } = Context.full_index () in
-    let rw2 = clone ~readonly:false in
+    let rw2 = clone ~readonly:false () in
     check_equivalence rw tbl;
     Index.close rw;
     Index.close rw2
 
   let find_absent () =
     let Context.{ rw; tbl; clone } = Context.full_index () in
-    let rw2 = clone ~readonly:false in
+    let rw2 = clone ~readonly:false () in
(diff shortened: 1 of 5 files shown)


test/unix/main.ml
```diff
@@ -332,6 +332,21 @@ module Close = struct
     Alcotest.check_raises "flush after double close with raises Closed"
       I.Closed (fun () -> Index.flush rw)
 
+  let restart_twice () =
+    let reuse_name = Context.fresh_name "empty_index" in
+    let rw = Index.v ~fresh:true ~readonly:false ~log_size:4 reuse_name in
+    let k1, v1 = (Key.v (), Value.v ()) in
+    Index.replace rw k1 v1;
+    Index.close rw;
+    let rw = Index.v ~fresh:true ~readonly:false ~log_size:4 reuse_name in
```
CraigFe: Could add an optional `~fresh` parameter to the `Context.t.clone` function to avoid building indexes yourself here.

test/unix/main.ml
```diff
@@ -332,6 +332,21 @@ module Close = struct
...
+    Alcotest.check_raises "Index restarts fresh cannot read data." Not_found
+      (fun () -> ignore (Index.find rw k1));
```
CraigFe: It's paranoid, but you could define an `ignore_value` utility function with a constrained type to avoid accidentally ignoring a partially-applied function if the `find` type ever changes:
```ocaml
let ignore_val (_ : value) = () in
```

icristescu: Nice trick! Thanks!

PR #136, opened 2019-11-28
Load log_async on startup

pascutto:
`log_async` was not loaded on startup, so here is a patch for this.
instead of loading it (which might cause issues on the next merge because it wouldn't be empty), I copied its data into the `log`.
The code probably requires some factoring, this is only a rough hotfix.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -324,6 +324,26 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
         iter_io (fun e -> Tbl.replace mem e.key e.value) io;
         Some { io; mem }
     in
+    let log_async_path = log_async_path root in
+    if (not readonly) && (not fresh) && Sys.file_exists log_async_path then
+      may
+        (fun log ->
+          let io =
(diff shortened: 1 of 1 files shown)


src/index.ml
```diff
@@ -324,6 +324,26 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
         iter_io (fun e -> Tbl.replace mem e.key e.value) io;
         Some { io; mem }
     in
+    let log_async_path = log_async_path root in
+    if (not readonly) && (not fresh) && Sys.file_exists log_async_path then
+      may
+        (fun log ->
+          let io =
+            IO.v ~fresh ~readonly ~generation:0L ~fan_size:0L log_async_path
```
CraigFe: Can you explicitly pass `~fresh:false ~readonly:false` here? I'm not sure _why_ I want it, but I do :slightly_smiling_face:

pascutto: I put it for `readonly`, I have a different handling of `fresh` in the last commit to still clean the file when `fresh` is `true`

PR #140, opened 2019-12-03
Specify directory for benchmarks

icristescu:


bench/db_bench.ml
```diff
@@ -118,8 +123,9 @@ module Index = struct
           ratio_reads)
 
   let init () =
-    if Sys.file_exists root then (
-      let cmd = Printf.sprintf "rm -rf %s" root in
+    (match !path with None -> () | Some path -> root := path // "db_bench");
```
pascutto: Is there a specific reason you need `path` as an intermediate?
Cmdliner could set the `root` directly, right?

PR #141, opened 2019-12-04
Refactor RO read

icristescu:
When reloading the `log_async`, there is no need to close and reopen the file, as it is not renamed during a merge. This is one of the causes of the difference in performance between a RW read and a RO read. 

`log_async` is reload when the offset of the file changed (due to an `add` or a `clear`) or when the generation changed, and new contents have potentially been added.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -401,37 +401,28 @@ module Make_private (K : Key) (V : Value) (IO : IO) = struct
           l "[%s] no changes detected" (Filename.basename t.root))
     in
     let add_log_entry log e = Tbl.replace log.mem e.key e.value in
+    let sync_log_async ?(generation_change = false) () =
+      match t.log_async with
+      | None -> t.log_async <- try_load_log t (log_async_path t.root)
+      | Some log ->
+          let offset = IO.offset log.io in
+          let new_offset = IO.force_offset log.io in
+          if generation_change || offset <> new_offset then (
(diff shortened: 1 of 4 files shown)


icristescu:
> What happens in the case here there is a `log`, but the generation hashn't changed? I feel like we used to read `log_async` in that case, but not any more with those changes, am I right ?

Oh yes, I forgot a case indeed.

icristescu:
Rebased over https://github.com/mirage/index/pull/144 to add the tests

src/index.ml
```diff
@@ -420,40 +420,33 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
...
     let add_log_entry log e = Tbl.replace log.mem e.key e.value in
+    let sync_log_async ?(generation_change = false) () =
+      match t.log_async with
+      | None -> t.log_async <- try_load_log t (log_async_path t.root)
+      | Some log ->
+          let offset = IO.offset log.io in
+          let new_offset = IO.force_offset log.io in
+          if generation_change || offset <> new_offset then (
+            Tbl.clear log.mem;
+            iter_io (add_log_entry log) ~max:new_offset log.io )
```
pascutto: There should be no need to specify `max` here, `iter_io` will use `IO.offset` anyway, which will return `new_offset` (it's fine to leave it though).

PR #142, opened 2019-12-04
Make transitive dependencies explicit in dune files

CraigFe:


CraigFe:
Failing because of some error with `format4` being abstract pre-4.07. 

```
File "bench/common.ml", line 9, characters 41-56:
Error: This expression has type ('a, 'b, 'c, 'd, 'e, 'f) format6
       but an expression was expected of type
         ('g, Format.formatter, unit, 'h) Stdlib.format4
```


I have no idea how to fix this, so unless someone has bright ideas I'll close until we drop support for 4.06.

avsm:
This seems like a bug with adding `stdlib` to the dune deps. Worth a bug upstream in dune to find out the right way to activate this mode.

PR #143, opened 2019-12-05
Add the option to write stats as a json output.

gs0510:
Depends on #134 :D

### DIFF
--- a/bench/db_bench.ml
+++ b/bench/db_bench.ml
@@ -69,7 +69,7 @@ let populate () =
   in
   loop 0
 
-let print_results db f nb_entries =
+let print_results db f nb_entries operation =
   let _, time = with_timer f in
   let micros = time *. 1_000_000. in
   let sec_op = micros /. float_of_int nb_entries in
@@ -77,7 +77,15 @@ let print_results db f nb_entries =
   let ops_sec = float_of_int nb_entries /. time in
   Log.app (fun l ->
       l "%s: %f micros/op; \t %f op/s; \t %f MB/s; \t total time = %fs." db
-        sec_op ops_sec mb time)
+        sec_op ops_sec mb time);
+  ( operation,
+    `Assoc
+      [
+        ("microsec_per_op", `Float micros);
+        ("ops_per_sec", `Float sec_op);
+        ("mb_per_sec", `Float mb);
+        ("total_time", `Float ops_sec);
+      ] )
 
 let rec random_new_key ar =
   let k = Context.Key.v () in
@@ -122,11 +130,11 @@ module Index = struct
 
   let bench_list =
     [
-      "fill_random";
-      "fill_seq";
-      "fill_seq_hash";
-      "fill_rev_seq_hash";
-      "fill_sync";
+      "write_random";
+      "write_seq";
+      "write_seq_hash";
+      "write_rev_seq_hash";
+      "write_sync";
     ]
 
   let namespace_v ?(fresh = true) ?(readonly = false) bench_name =
@@ -135,9 +143,9 @@ module Index = struct
   let write_bench ~bench bench_name =
     if List.mem bench_name bench_list then (
       Stats.reset_stats ();
-      let rw = bench (namespace_v bench_name) in
+      let rw, output = bench (namespace_v bench_name) in
       write_amplif ();
-      rw )
+      (rw, output) )
     else failwith "Add the bench name to bench_list"
 
   let cleanup () =
@@ -173,31 +181,34 @@ module Index = struct
         add_metrics ())
       random
 
-  let write_random () =
+  let write_random name =
     let bench rw =
-      print_results (write rw) nb_entries;
-      rw
+      let output = print_results (write rw) nb_entries name in
+      (rw, output)
     in
-    write_bench ~bench "fill_random"
+    let rw, output = write_bench ~bench name in
+    (rw, [ output ])
 
-  let write_seq () =
+  let write_seq name =
     let bench rw =
       Array.sort (fun a b -> String.compare (fst a) (fst b)) random;
-      print_results (write rw) nb_entries;
-      Index.close rw
+      let output = print_results (write rw) nb_entries name in
+      Index.close rw;
+      ((), output)
     in
-    ignore (write_bench ~bench "fill_seq")
+    [ snd (write_bench ~bench name) ]
 
-  let write_seq_hash () =
+  let write_seq_hash name =
     let bench rw =
       let hash e = Context.Key.hash (fst e) in
       Array.sort (fun a b -> compare (hash a) (hash b)) random;
-      print_results (write rw) nb_entries;
-      Index.close rw
+      let output = print_results (write rw) nb_entries name in
+      Index.close rw;
+      ((), output)
     in
-    ignore (write_bench ~bench "fill_seq_hash")
+    [ snd (write_bench ~bench name) ]
 
-  let write_rev_seq_hash () =
+  let write_rev_seq_hash name =
     let bench rw =
       let hash e = Context.Key.hash (fst e) in
       Array.sort (fun a b -> compare (hash a) (hash b)) random;
@@ -208,12 +219,13 @@ module Index = struct
             add_metrics ())
           random
       in
-      print_results (write rw) nb_entries;
-      Index.close rw
+      let output = print_results (write rw) nb_entries name in
+      Index.close rw;
+      ((), output)
     in
-    ignore (write_bench ~bench "fill_rev_seq_hash")
+    [ snd (write_bench ~bench name) ]
 
-  let write_sync () =
+  let write_sync name =
     let bench rw =
       let write rw () =
         Array.iter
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ocaml/opam2
+RUN sudo apt install -y m4 perl libgmp-dev pkg-config
+COPY  --chown=opam:opam  .  index
```
samoht: you might want to just copy `*.opam` instead of `.`, that will cache the dependencies better.

bench/db_bench.ml
```diff
@@ -35,7 +17,7 @@ let hash_size = 30
 
 let value_size = 13
 
-let nb_entries = 10_000_000
+let nb_entries = 10_000
```
samoht: is that change necessary? Maybe we want that number to be passed on the command line to ease tweaking?

gs0510: Whoops, this should not be changed. Also this is from #134 (I'll pull in that branch once it gets merged)

gs0510:
@icristescu @CraigFe

bench/db_bench.ml
```diff
@@ -400,8 +419,13 @@ let input =
   in
   Arg.(value & opt options `Minimal & info [ "b"; "bench" ] ~doc)
 
+let json_output =
+  let doc = "Filename where json output should be written" in
+  Arg.(value & opt string "" & info [ "j"; "json" ] ~doc)
```
```suggestion
  Arg.(value & opt (some non_dir_file) None & info [ "j"; "json" ] ~doc)
```
I think such null values should be exposed in the type. You could make `json_file` a `string option` with something like this. Also, Cmdliner has a `non_dir_file` Arg that will do a bit of validation for you (and improve the `--help` docs).

bench/main.ml
```diff
@@ -90,8 +101,13 @@ let input =
   in
   Arg.(value & opt options `All & info [ "b"; "bench" ] ~doc)
 
+let json_output =
```
CraigFe: Should probably this in a `cli.ml` file to avoid duplication from `db_bench`. We can do a later refactoring pass to ensure that the two Cmdliner APIs for `main` and `db_bench` are fully consistent.

bench/main.ml
```diff
@@ -44,33 +44,44 @@ let rec finds t count = function
...
     | `Find `RO | `All ->
         let ro = Index.v ~readonly:true ~log_size index_name in
         let (), t3 = with_timer (fun () -> finds ro 0 bindings) in
         Index.close ro;
-        Fmt.epr "\n%d bindings found in %fs (RO).\n%!" index_size t3
-    | _ -> ()
+        Fmt.epr "\n%d bindings found in %fs (RO).\n%!" index_size t3;
+        ("read_only", `Float t3) :: output_json
+    | _ -> output_json
+  in
```
CraigFe: Ouch. I didn't know we had this in here. This should probably be refactored to be the same as in `db_bench.ml`, but we can do that in a later PR.

bench/db_bench.ml
```diff
@@ -35,6 +35,8 @@ let entry_size = key_size + value_size
 
 let random = Array.make nb_entries ("", "")
 
+let json_output = ref []
```
CraigFe: It's generally considered bad practice to have global state like this. Perhaps you could have each benchmark return its results directly, then `List.map` rather than `List.iter` to get those results.

bench/db_bench.ml
```diff
@@ -113,15 +131,15 @@ module Index = struct
...
+    print_results (write rw) nb_entries "random_writes";
     write_amplif ();
     rw
 
   let write_seq () =
     Index_unix.reset_stats ();
     let rw = Index.v ~fresh:true ~log_size (root // "fill_seq") in
     Array.sort (fun a b -> String.compare (fst a) (fst b)) random;
-    print_results (write rw) nb_entries;
+    print_results (write rw) nb_entries "sequential_writes";
```
CraigFe: We now have 3 strings associated with almost all benchmarks:
 1. the name of the output directory for data (`"fill_seq"` here)
 2. the key of the JSON output data (`"sequential_writes"` here)
 3. the name of the benchmark in the CLI (`"write-keys"` here)

These are often not the same, as in this case, which strikes me as quite confusing. We could unify at least 1 & 2 by passing a `name` string as input to each benchmark. 

3 is not related to this PR, and can be caught later on :slightly_smiling_face:

bench/db_bench.ml
```diff
@@ -46,15 +46,29 @@ let populate () =
   in
   loop 0
 
-let print_results db f nb_entries =
+let print_results db f nb_entries operation =
```
icristescu: maybe rename this, it's no longer printing, but it returns a Json. 
also I think we don't need both `json` and `Log.app`. what do you think of letting the user decide which type of output it wants? for instance, if no json file is provided then only print to terminal ?

bench/db_bench.ml
```diff
@@ -110,91 +124,88 @@ module Index = struct
...
 
-  let write_random () =
+  let write_random name =
     Index_unix.reset_stats ();
-    let rw = Index.v ~fresh:true ~log_size (root // "fill_random") in
-    print_results (write rw) nb_entries;
+    let rw = Index.v ~fresh:true ~log_size (root // name) in
     write_amplif ();
-    rw
+    (rw, [ print_results (write rw) nb_entries name ])
```
icristescu: I think you don't need to return a list here, but directly the results of `print_results`, since you always return a list of one element. This way you don't need to call `list.flatten` at the end either

bench/db_bench.ml
```diff
@@ -110,91 +124,88 @@ module Index = struct
 
   let read r () = Array.iter (fun (k, _) -> ignore (Index.find r k)) random
 
-  let write_random () =
+  let write_random name =
     Index_unix.reset_stats ();
-    let rw = Index.v ~fresh:true ~log_size (root // "fill_random") in
-    print_results (write rw) nb_entries;
```
icristescu: you can't reverse the order of `print_results` and `write_amplif` because `write_amplif` uses the stats gather during  the execution of the bench in `print_results`. I almost didn't see this ! Could you run the benchmarks before and after your changes to check that everything looks roughly the same?

bench/db_bench.ml
```diff
@@ -323,51 +339,60 @@ let init () =
...
   let match_input ~bench ~triggers ~message =
     if List.mem input triggers then
       let () = Log.app (fun l -> l "\n %s" message) in
       bench ()
-    else ()
+    else []
+  in
+  let output_json =
+    List.fold_left
+      (fun foo (bench, triggers, message) ->
```
icristescu: personal preferences here, but rename `foo` in `acc` or `output_json` or something suggestive

bench/db_bench.ml
icristescu: as @CraigFe suggested :)

bench/main.ml
```diff
@@ -44,33 +44,44 @@ let rec finds t count = function
       ignore (Index.find t k);
       finds t (count + 1) tl
 
-let run input =
+let print_json input file =
```
icristescu: you can maybe move this to `common`? as it is used in both `db_bench.ml` and `main.ml`

gs0510:
I moved the write_json function to common but the I had to also change main.ml functions to have [[]] this format :crying_cat_face: There should be a better way.

Also, why does Travis CI fail? :fearful: I have dune in index.opam

pascutto:
The CI fails because `yojson` is specified in `index-bench.opam`, but the build is triggered by `index` tests, where `yojson` is not a dependency. That's a problem in `bench/dune` from the time when we did not have `index-bench.opam`.

bench/db_bench.ml
```diff
@@ -139,43 +147,47 @@ module Index = struct
...
+  let write_seq name =
     let bench rw =
       Array.sort (fun a b -> String.compare (fst a) (fst b)) random;
-      print_results (write rw) nb_entries;
-      Index.close rw
+      let output = print_results (write rw) nb_entries name in
+      (Index.close rw, output)
     in
-    ignore (write_bench ~bench "fill_seq")
+    let _, output = write_bench ~bench name in
```
icristescu: you can also write this using `snd (write_bench ...)`

bench/db_bench.ml
```diff
@@ -291,12 +312,12 @@ module Lmdb = struct
               Lmdb.put_string txn ddb k v >>= fun () -> sync env))
         ls
     in
-    print_results (write rw env random) nb_entries;
-    closedir env
+    closedir env;
```
icristescu: Here you can't reverse the two either, as `print_results (write rw env ...)` uses `env`

icristescu:
> I moved the write_json function to common but the I had to also change main.ml functions to have [[]] this format 😿 There should be a better way.

I agree, it is annoying. Did you try to use `(List of json list) : json` instead of json list? maybe we can pass around json values instead of json lists.

bench/db_bench.ml
```diff
@@ -48,15 +48,23 @@ let populate () =
...
   let _, time = with_timer f in
   let micros = time *. 1_000_000. in
   let sec_op = micros /. float_of_int nb_entries in
   let mb = float_of_int (entry_size * nb_entries / 1_000_000) /. time in
   let ops_sec = float_of_int nb_entries /. time in
   Log.app (fun l ->
-      l "%s: %f micros/op; \t %f op/s; \t %f MB/s; \t total time = %fs." db
-        sec_op ops_sec mb time)
+      l "%s %s: %f micros/op; \t %f op/s; \t %f MB/s; \t total time = %fs." db
+        operation sec_op ops_sec mb time);
```
icristescu: I'm not sure if you should print `operation` here as is something like `fill_random`, not very intelligible as a user output. Also we print the operation before doing it anyway.

gs0510:
> > I moved the write_json function to common but the I had to also change main.ml functions to have [[]] this format crying_cat_face There should be a better way.
> 
> I agree, it is annoying. Did you try to use `(List of json list) : json` instead of json list? maybe we can pass around json values instead of json lists.
> 
> For instance you can have _something like this_
> 
> ```ocaml
>       match accu with
>         | `List acc -> (
>             match (match_input ~bench ~triggers ~message) with
>             | `List ls -> `List (ls @ acc)
>             | `Assoc _ as j2 -> `List (j2 :: acc)
>             | _ -> accu )
>         | _ -> accu)
>       (`List [ output_json ])
> ```
> 
> if you pass around Json-structure values then you can decompose them when needed it. It allows you to have the return type of match_input both a list and an assoc.
> 
> Otherwise, we can also treat the lmdb case separately.

I am going to do this in a separate refactoring PR (Since that would involve a lot of changes to this anyway)

icristescu:
> I moved the write_json function to common but the I had to also change main.ml functions to have [[]] this format 😿 There should be a better way.

Maybe you can call `List.flatten` before calling `write_json` in `db_bench`?

PR #144, opened 2019-12-09
Wait for merge thread to finish in tests

icristescu:
In order to prevent weird debug messages during `dune runtest`, we can wait for the merge threads to finish, before exiting a test. 

However, in `index` we do not want to wait for the merge thread to finish, and therefore some debug messages can still appear during a runtest. 

There is an issues in this PR which is that the test `test_fail` does not fail and it should.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -691,7 +691,7 @@ module Make_private (K : Key) (V : Value) (IO : IO) = struct
           Int64.compare (IO.offset log.io) (Int64.of_int t.config.log_size) > 0)
     in
     if do_merge then
-      ignore (merge ~witness:{ key; key_hash = K.hash key; value } t)
+      ignore (merge ~witness:{ key; key_hash = K.hash key; value } t : async)
 
   let iter f t =
     let t = check_open t in
--- a/src/index.ml
+++ b/src/index.ml
@@ -84,6 +84,10 @@ exception RO_not_allowed
 exception Closed
 
 module Make_private (K : Key) (V : Value) (IO : IO) = struct
+  type async = IO.async
+
+  let await = IO.await
+
   type key = K.t
 
   type value = V.t
@@ -666,7 +670,8 @@ module Make_private (K : Key) (V : Value) (IO : IO) = struct
     let witness = IO.Mutex.with_lock t.rename_lock (fun () -> get_witness t) in
     match witness with
     | None ->
-        Log.debug (fun l -> l "[%s] index is empty" (Filename.basename t.root))
+        Log.debug (fun l -> l "[%s] index is empty" (Filename.basename t.root));
+        IO.return ()
     | Some witness -> merge ?hook ~witness t
 
   let replace t key value =
@@ -685,7 +690,8 @@ module Make_private (K : Key) (V : Value) (IO : IO) = struct
           Tbl.replace log.mem key value;
           Int64.compare (IO.offset log.io) (Int64.of_int t.config.log_size) > 0)
     in
-    if do_merge then merge ~witness:{ key; key_hash = K.hash key; value } t
+    if do_merge then
+      ignore (merge ~witness:{ key; key_hash = K.hash key; value } t)
 
   let iter f t =
     let t = check_open t in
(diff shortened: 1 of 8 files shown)


icristescu:
I left `test_fail` in the first commit to have an example of this in the history.

src/index.ml
```diff
@@ -82,7 +82,12 @@ module type S = sig
 
   val iter : (key -> value -> unit) -> t -> unit
 
-  val force_merge : ?hook:[ `After | `Before ] Private.Hook.t -> t -> unit
+  type async_type
+
+  val force_merge :
+    ?hook:[ `After | `Before ] Private.Hook.t -> t -> async_type option
+
+  val join : async_type option -> unit
```
pascutto: I'm a bit reluctant to having this in the public API, maybe it could be moved to `Private`?

CraigFe: Do you not like the notion of being able to bind on the `force_merge` finishing at all, or just think it's a bit ugly? I think it might be prettier as:
```ocaml
type async

val force_merge : ?hook:[ `After | `Before ] Private.Hook.t -> t -> unit async
val await : 'a async -> 'a
```

CraigFe: (or switch `async`/`await` to `promise`/`resolve` if you're the web-developing type)

icristescu: @pascutto I leave the `Private` part to https://github.com/mirage/index/pull/147 then 

@CraigFe you are right, I can hide the option in the type `async` as well (done now), but I'm not sure why we should have a parametric type here.

CraigFe: No good reason :slightly_smiling_face: I suggested it because to me the type `'a t -> 'a` strongly suggests the implementation "_block on the resolution of 'a-valued promise_", by analogy to:

|    |   |   |
| -- | -- | -- |
| OCaml | [`Lwt_main.run`](https://ocsigen.org/lwt/4.4.0/api/Lwt_main) | `'a t -> 'a` |
| Haskell | [`wait`](http://hackage.haskell.org/package/async-2.2.2/docs/Control-Concurrent-Async.html#v:wait) | `Async a -> IO a` |
| Rust  | [`join`](https://doc.rust-lang.org/std/thread/struct.JoinHandle.html#methods) | `JoinHandle<A> -> Result<A>` |

whereas `t -> unit` doesn't speak to me as much (e.g. it might do cancellation like `Lwt.cancel : 'a t -> unit` or `cancel :: Async a -> IO ()`). But on closer inspection our Unix threads are old-fashioned value-less ones anyway :stuck_out_tongue: Sorry!

CraigFe: (it also makes effectively no difference if the type is private anyway...)

src/io.mli
```diff
@@ -76,5 +76,11 @@ module type S = sig
     val with_lock : t -> (unit -> 'a) -> 'a
   end
 
-  val async : (unit -> 'a) -> unit
+  type async_type
```
CraigFe:
```suggestion
  type async
```
I guess we rename this `async` to match now?

src/index.ml
```diff
@@ -701,7 +710,8 @@ module Make (K : Key) (V : Value) (IO : IO) = struct
           Tbl.replace log.mem key value;
           Int64.compare (IO.offset log.io) (Int64.of_int t.config.log_size) > 0)
     in
-    if do_merge then merge ~witness:{ key; key_hash = K.hash key; value } t
+    if do_merge then
+      ignore (merge ~witness:{ key; key_hash = K.hash key; value } t)
```
```suggestion
      ignore (merge ~witness:{ key; key_hash = K.hash key; value } t : async)
```

src/index.ml
icristescu: I agree that the type `'a t ->'a` is more suggestive. I propose we leave like this for now, and the PR that uses Lwt will change a lot of things anyway and we can change this too then.

pascutto:
You need to rebase to fix the CI

icristescu:
> You need to rebase to fix the CI

Ocaml-ci still fails (it does in the main code too so it's not because of the PR).

pascutto:
The remaining error is a known issue about system dependencies updates on ubuntu, we can ignore it.

PR #145, opened 2019-12-09
Refactor the clean up operation in db_bench

icristescu:
So as to prevent erasing other files in the `db_bench` directory.

### DIFF
--- a/bench/db_bench.ml
+++ b/bench/db_bench.ml
@@ -99,66 +99,95 @@ module Index = struct
         l "\tread amplification in bytes = %f; in nb of reads = %f " ratio_bytes
           ratio_reads)
 
-  let init () =
-    if Sys.file_exists root then (
-      let cmd = Printf.sprintf "rm -rf %s" root in
-      Logs.app (fun l -> l "exec: %s\n" cmd);
-      let _ = Sys.command cmd in
-      () )
+  let bench_list =
+    [
+      "fill_random";
+      "fill_seq";
+      "fill_seq_hash";
+      "fill_rev_seq_hash";
+      "fill_sync";
+    ]
+
+  let namespace_v ?(fresh = true) ?(readonly = false) bench_name =
+    Index.v ~fresh ~log_size ~readonly (root // bench_name)
+
+  let write_bench ~bench bench_name =
+    if List.mem bench_name bench_list then (
+      Index_unix.reset_stats ();
+      let rw = bench (namespace_v bench_name) in
+      write_amplif ();
+      rw )
+    else failwith "Add the bench name to bench_list"
+
+  let cleanup () =
+    let files = [ "data"; "log"; "lock"; "log_async"; "merge" ] in
+    List.iter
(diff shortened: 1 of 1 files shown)


bench/db_bench.ml
```diff
@@ -77,6 +77,14 @@ module Index = struct
 
   let root = "_bench" // "db_bench"
 
+  let names =
+    let c = ref 0 in
+    fun object_type ->
```
CraigFe: AFAICS, the arguments to this function aren't really `object_type`s. Could we use it as the object name directly? (rather than creating less-useful `index_#` names)

bench/db_bench.ml
```diff
@@ -99,35 +107,47 @@ module Index = struct
...
+        (fun file ->
+          let file = dir // file in
+          if Sys.file_exists file then (
+            let cmd = Printf.sprintf "rm %s" file in
+            Logs.app (fun l -> l "exec: %s\n" cmd);
+            let _ = Sys.command cmd in
+            () ))
+        files
+    in
+    for i = 0 to 4 do
```
```suggestion
    for i = 0 to 4 do
```
Suspicious magic number 4. Does this come from the fact that you know `names` is called 5 times?

bench/db_bench.ml
```diff
@@ -327,13 +345,13 @@ let run input =
...
   Log.app (fun l -> l "Fill in random order");
-  let rw = Index.write_random () in
+  let rw, name = Index.write_random () in
   let bench_list =
     [
       ( (fun () -> Index.read_random rw),
         [ `Read `RW; `All; `Minimal; `Index ],
         "RW Read in random order" );
-      ( (fun () -> Index.ro_read_random rw),
+      ( (fun () -> Index.ro_read_random rw name),
```
CraigFe: If this benchmark needs a particular index instance `name`, could we just pass it directly here rather than sharing it's path and reconstructing?

bench/db_bench.ml
icristescu: yes :) I know it's not great, I don't see a nice way to refactor this (as cleanup is called at the beginning of a run).
I can instead have a list of all possible index names used in the benchmarks, maybe that is cleaner?

CraigFe: This is perhaps over-complicating things, but we could keep the benchmark names outside in the `bench_list` variable, then have each benchmark consume a constructor namespaced with that string:
```ocaml
let namespace_v bench_name ~fresh ~log_size ~readonly name =
  Index.v ~fresh ~log_size ~readonly (root // bench_name // name)

...

let match_input ~bench ~triggers ~bench_name =
  if List.mem input triggers then
    let () = Log.app (fun l -> l "\n %s" bench_name) in
    bench (namespace_v ~bench_name)
```

This way you can clean by iterating over `bench_list` and purging all of the `root // bench_name` directories.

icristescu: Thanks ! I changed it as you proposed, and I think it is better.

PR #146, opened 2019-12-11
Build benchs with index-bench instead of index

pascutto:
This should fix the issue described here: https://github.com/mirage/index/pull/143#issuecomment-564592608
Also, that would check that `db_bench` compiles correctly in the CI.

### DIFF
--- a/bench/dune
+++ b/bench/dune
@@ -1,22 +1,21 @@
 (executable
- (name main)
+ (public_name main)
  (modules main)
+ (package index-bench)
  (libraries fmt index.unix common cmdliner))
 
 (executable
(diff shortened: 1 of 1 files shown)


bench/dune
```diff
@@ -1,23 +1,16 @@
...
+ (public_name db_bench)
  (modules db_bench)
+ (package index-bench)
  (libraries fmt index.unix lmdb common bigstring rresult logs logs.fmt
    cmdliner))
 
-(alias
- (name bench)
- (action
-  (run ./main.exe)))
```
CraigFe: We probably still want this.

PR #147, opened 2019-12-11
Extract a distinct private API for Index

CraigFe:
This splits out a separate API for Index to be used under test. For the moment, the only difference between the two is that `force_merge` is unavailable in the public API: this will allow us to properly test concurrency properties that we don't want to make available to users (see https://github.com/mirage/index/pull/144#discussion_r356645024).

We could also choose to hide other functions, such as `iter`, which we can't give very sensible semantics to currently. I'm interested to know your opinions on this.

### DIFF
--- a/src/index.mli
+++ b/src/index.mli
@@ -161,7 +161,11 @@ module Private : sig
     include S
 
     val force_merge : ?hook:[ `After | `Before ] Hook.t -> t -> unit
-    (** [force_merge t] forces a merge for [t]. *)
+    (** [force_merge t] forces a merge for [t]. Optionally, a hook can be passed
+        that will be called twice:
+
+        - [`Before]: immediately before merging (while holding the merge lock);
+        - [`After]: immediately after merging (while holding the merge lock). *)
   end
 
   module Make (K : Key) (V : Value) (IO : IO) :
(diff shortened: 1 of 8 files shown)


src/index.mli
```diff
@@ -158,3 +139,29 @@ end
...
+
+  module Io_array : module type of Io_array
+
+  module Fan : module type of Fan
+
+  module type S = sig
+    include S
+
+    val force_merge : ?hook:[ `After | `Before ] Hook.t -> t -> unit
+    (** [force_merge t] forces a merge for [t]. *)
```
pascutto: Not strictly related to this PR, but we could take advantage of it to specify a bit more clearly how to use the hooks.

CraigFe: Indeed: done in https://github.com/mirage/index/pull/147/commits/6e1395c0de735414f56cd0e3b21a3e8687fef477.

CraigFe:
I've moved [iter] to the Private API too. This necessitates the benchmarks using the Private API, which seems fair as they are already relying on an implicit understanding of how merges behave in Index.

pascutto:
That's ok indeed, https://github.com/mirage/index/pull/115 will need to access `force_merge` thus will use `Private` too.

src/index.mli
pascutto: Looks good, maybe just specify that it happens in the child thread?

CraigFe:
Now rebased over https://github.com/mirage/index/pull/100.

PR #150, opened 2019-12-16
Revert "Move Index.iter to Index.Private.iter"

CraigFe:
This reverts commit b82dbf3da95c752096e6b174c66cbed7262d7233. The `iter`
API endpoint is relied upon by Irmin in order to do integrity checking,
so it can't be moved to the private API. See
https://github.com/mirage/irmin/pull/904.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,9 +2,8 @@
 
 ## Removed
 
- - Remove `iter` and `force_merge` from `Index.S`, due to difficulties with
-   guaranteeing sensible semantics for these functions under MRSW access
-   patterns. (#147)
+ - Remove `force_merge` from `Index.S`, due to difficulties with guaranteeing
+   sensible semantics to this function under MRSW access patterns. (#147, #150)
 
 # 1.0.1 (2019-11-29)
 
--- a/src/index.ml
+++ b/src/index.ml
@@ -70,6 +70,8 @@ module type S = sig
 
   val replace : t -> key -> value -> unit
 
+  val iter : (key -> value -> unit) -> t -> unit
+
   val flush : t -> unit
 
   val close : t -> unit
@@ -739,8 +741,6 @@ module Private = struct
 
   module type S = sig
     include S
-
-    val iter : (key -> value -> unit) -> t -> unit
 
     type async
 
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -2,9 +2,8 @@
 
 ## Removed
 
- - Remove `iter` and `force_merge` from `Index.S`, due to difficulties with
-   guaranteeing sensible semantics for these functions under MRSW access
-   patterns. (#147)
+ - Remove `force_merge` from `Index.S`, due to difficulties with guaranteeing
```
icristescu: The difficulties mentioned here are related to `iter`, no?

pascutto:
Shouldn't we fix it (or at least think about how it's broken and provide a clear specification) before exposing it again?

CraigFe:
The last time we looked at this, we decided that it was impossible (or at the least, very hard) to ascribe sensible semantics to this with the locks we currently have available. I'm happy to update the doc comment to give a clearer warning about the guarantees that we're _not_ providing, but it's not even clear to me what those are :slightly_smiling_face:

CHANGES.md
CraigFe: They're still related to `force_merge`, I think. If we need to support `iter`, then I suppose the best we can do is give a big warning in the documentation comment as @pascutto suggests.

icristescu:
> The last time we looked at this, we decided that it was impossible (or at the least, very hard) to ascribe sensible semantics to this with the locks we currently have available. I'm happy to update the doc comment to give a clearer warning about the guarantees that we're _not_ providing, but it's not even clear to me what those are 🙂

We cannot guarantee a non-duplicates iter for RO https://github.com/mirage/index/pull/116. Because RO cannot do a merge before an iter, duplicate values can exists in log and index. But if we don't ask for non-duplicates iter I think we are fine.

pascutto:
We cannot guarantee that all the entries are hit either. For instance in `find` we retry `sync_log` because we can somehow miss an update (and items in it), we don't do that in `iter` which means not reaching some bindings is plausible.
Also, the use of the lock is suspicious since this lock will be useless for multiprocess use cases, so that may also lead to concurrency issues and bindings not hit.
I think the current implementation should really not be part of the public API without a strong warning until we fix it somehow.

CraigFe:
Rebased and added documentation comments. Will merge as soon as the CI is happy.

PR #151, opened 2019-12-16
Benchmarks refactoring

pascutto:
- `main.ml` and `db_bench.ml` benchmarks are merged into a single file (and common is not needed anymore).
- Benchmarks that depend on each other now do in an explicit way and the order of execution is dynamically computed.
- Printing is now done in only one place, allowing to easily switch between json and human mode.
- Cmdliner uses the benchmark definition directly instead of having a separate configuration
- command line is improved  by adding options and regexp parsing

I'm sorry this is probably hard to review as a diff, I could not really make it better

I did not implement the lmdb benchmarks yet, let me know if they are still relevant for the comparison and we can add them back.


bench/bench.ml
```diff
@@ -0,0 +1,480 @@
...
+      result.write_amplification_size
+end
+
+let make_bindings_pool nb_entries =
+  Array.init nb_entries (fun _ ->
+      let k = Context.Key.v () in
+      let v = Context.Value.v () in
+      (k, v))
+
+let bindings_pool = ref [||]
```
icristescu: Why do you have `ref` for these arrays?

bench/bench.ml
```diff
@@ -0,0 +1,480 @@
...
+
+  let read_absent ~with_metrics bindings r =
+    Array.iter
+      (fun (k, _) ->
+        try ignore (Index.find r k : Context.Value.t)
+        with Not_found -> if with_metrics then add_metrics ())
+      bindings
+
+  let write_random ~with_metrics v () =
+    v ~fresh:true ~readonly:false "" |> write ~with_metrics !bindings_pool
```
icristescu: Does it mean that the benchmarks also measures the time to create/open the index? I think we could add a setup function for each benchmark, to run before? What do you think?

bench/bench.ml
```diff
@@ -0,0 +1,480 @@
...
+      nb_entries:int ->
+      log_size:int ->
+      root:string ->
+      name:string ->
+      benchmark ->
+      Benchmark.result =
+   fun ~with_metrics ~nb_entries ~log_size ~root ~name b ->
+    let indices = ref [] in
+    let index_v ~fresh ~readonly n =
+      let i = Index.v ~fresh ~readonly ~log_size (root // name // n) in
```
icristescu: `n` seems to be "" for all benchmarks, no?

bench/bench.ml
```diff
@@ -0,0 +1,480 @@
...
+let schedule p s =
+  let todos = List.map fst in
+  let init = ref (s |> List.map (fun b -> (p b.Index.name, b))) in
+  let apply_dep s =
+    let deps =
+      s
+      |> List.fold_left
+           (fun acc (todo, b) ->
+             if todo then
+               match b.Index.dependency with Some s -> s :: acc | None -> acc
```
icristescu: You could do the same thing in `cleanup` to avoid cleaning directories that do not exist.

bench/bench.ml
```diff
@@ -0,0 +1,480 @@
...
+
+let init config =
+  Printexc.record_backtrace true;
+  Random.init config.seed;
+  if config.with_metrics then (
+    Metrics.enable_all ();
+    Metrics_gnuplot.set_reporter ();
+    Metrics_unix.monitor_gc 0.1 );
+  bindings_pool := make_bindings_pool config.nb_entries;
+  absent_bindings := make_bindings_pool config.nb_entries
```
icristescu: I think you need to check that the values you add in `absent_binding` are not in `binding_pool`.

icristescu:
Also, I think you forgot to add the possibility to choose which benchmarks to run, I think it would be nice to have it as well.

bench/bench.ml
pascutto: The goal is to save time on startup so that if you're asking for `bench.exe list` for instance, or `bench.exe --help`, then you get immediate output instead of having to wait for one minute for these to fill.
I agree having a `ref` for this is ugly though, I could probably pass these values instead of having them global.

icristescu: Indeed, I haven't realised that in the previous version, populating the array blocks these commands.

bench/bench.ml
pascutto: Actually the probability of getting the same key twice is neglectable, so I thought it's not worth the effort of checking since this has a significant cost.

icristescu: I think that it's ok if the setup for a benchmark takes a lot of time. I prefer having a smaller array of absent values if I want a fast setup, but ensuring that they are indeed absent. But that's just my personal preference :)

pascutto:
> Also, I think you forgot to add the possibility to choose which benchmarks to run, I think it would be nice to have it as well.

That is done in [here](https://github.com/mirage/index/pull/151/files#diff-e3759a50620e39708328f84ebc42287cR421-R428) and [here](https://github.com/mirage/index/pull/151/files#diff-e3759a50620e39708328f84ebc42287cR404-R406).

> I'm wondering about having common and bench merged.

That indeed makes since we want lmdb back, I'll restore that.

bench/bench.ml
pascutto: Indeed, will move it here instead!

CraigFe: Hmm, I'm not sure it _should_ be moved here. 

I think `index_v` should have the same type as `Index.v`; this is just a scoping and finalising mechanism for any indices that are constructed by the benchmarks. 

The fact that each benchmark only constructs a single index (and therefore puts it in `""`) feels like more of a coincidence than anything else. Shouldn't benchmarks be capable of constructing more than one index?

pascutto: That was indeed the motivation for leaking the name argument in my PR, but on the other hand I don't see what benchmark would need to create multiple indices (without being a combination of already existing benchmarks), so I don't mind removing it if that's easier to understand.
There is however an issue with the current implementation: `cleanup` only cleans the `""` index right now, but that's easy to fix.

bench/bench.ml
pascutto: Summary of the discussion we had offline:
1. The probability of having a common element in those two arrays is ~10^-49 for 10_000_000 elements and can go up to ~10^-45 for 1_000_000_000 elements so that is indeed neglectable (can provide the calculus but GitHub is not great at formulas).
2. The time needed *with verification* to generate a size 1000 array of absent values when the main pool is of size 10_000_000 (current behavior) is approximately 90 seconds.
3. The time needed *without verification* (this PR) to generate a 10_000_000 array of (unverified) absent values is 12 seconds; that would not be possible with the verification.
4. If we make a mistake by actually having a common element, the impact on the result would also be neglectable given the size of the dataset.

Given those points, it's probably not worth having the verification.

Side note to mitigate point 1: if we were to use a much smaller key size (<10) the probability would rise beyond neglectability, but I don't think we have a use case for keys that small.

bench/bench.ml
pascutto: I'll only allow benchs to create one index for now, we can fix it in another PR if needed.

bench/bench.ml
pascutto: `cleanup` will already check if a file exists before trying to remove it

bench/bench.ml
pascutto: They do indeed, I thought it was part of what we wanted to measure (btw we have no dedicated bench for `v`). Would you prefer to have the `fresh` and `readonly` parameters inside the record and the function called before the timer starts to run?

pascutto: They do indeed, I thought it was part of what we wanted to measure (btw we have no dedicated bench for v). Would you prefer to have the fresh and readonly parameters inside the record and the function called before the timer starts to run?

icristescu: We don't have a benchmark for `v` and we could, indeed! But I prefer that the benchmarks for write and read only measure the performance of write and read respectively and not the time the db sets up for those.

bench/bench.ml
```diff
@@ -0,0 +1,487 @@
...
+  let doc = "A regular expression matching the names of benchmarks to run" in
+  let env = env_var "NAME_FILTER" in
+  Arg.(
+    value
+    & opt (some regex) None
+    & info [ "f"; "filter" ] ~env ~doc ~docv:"NAME_REGEX")
+
+let data_dir =
+  let doc = "Set directory for the data files" in
+  let env = env_var "DATA_DIR" in
```
CraigFe:
```suggestion
  let env = env_var "INDEX_DATA_DIR" in
```
I prefer having some sort of prefix here, just in case these interface with e.g. Dune invocations.

bench/bench.ml
```diff
@@ -0,0 +1,487 @@
...
+  absent_bindings := make_bindings_pool config.nb_entries
+
+let print fmt (config, results) =
+  let pp_bench fmt (b, result) =
+    Format.fprintf fmt "%s@\n    @[%a@]" b.Index.synopsis Benchmark.pp_result
+      result
+  in
+  Format.fprintf fmt
+    "Configuration:@\n    @[%a@]@\n@\nResults:@\n    @[%a@]@\n" pp_config config
+    Fmt.(list ~sep:Fmt.(any "@\n@\n") pp_bench)
```
CraigFe:
```suggestion
    Fmt.(list ~sep:(any "@\n@\n") pp_bench)
```
This really feels like it should be a compiler warning :thinking:

bench/bench.ml
```diff
@@ -0,0 +1,487 @@
...
+  Arg.(value & opt int 500_000 & info [ "log_size" ] ~env ~doc)
+
+let nb_entries =
+  let doc = "The number of bindings." in
+  let env = env_var "NB_ENTRIES" in
+  Arg.(value & opt int 10_000_000 & info [ "nb_entries" ] ~env ~doc)
+
+let list_cmd =
+  let doc = "List all available benchmarks." in
+  (Term.(pure list_benchs $ const ()), Term.info "list" ~doc)
```
CraigFe:
```suggestion
  (Term.(pure list_benches $ const ()), Term.info "list" ~doc)
```

bench/bench.ml
pascutto: This is already prefixed with `INDEX_BENCH_` in the definition of `env_var`.

CraigFe: I see, OK.

bench/bench.ml
pascutto: Indeed.

bench/bench.ml
pascutto: Gave it a try in https://github.com/mirage/index/pull/151/commits/3fda3e5c6e0312478ad2e33003ccb23cac0d6323

bench/bench.ml
```diff
@@ -0,0 +1,484 @@
...
+    mbs_per_sec : float;
+    read_amplification_calls : float;
+    read_amplification_size : float;
+    write_amplification_calls : float;
+    write_amplification_size : float;
+  }
+  [@@deriving yojson]
+
+  let run ~nb_entries f =
+    let _, time, stats = with_stats (fun () -> f ()) in
```
icristescu: If this is the only place where you call `with_stats`, you don't need to return a triple then.

pascutto: Solved! :heavy_check_mark:

bench/bench.ml
```diff
@@ -0,0 +1,484 @@
...
+    Array.iter
+      (fun (k, _) ->
+        try ignore (Index.find r k : Context.Value.t)
+        with Not_found -> if with_metrics then add_metrics ())
+      bindings
+
+  let write_random ~with_metrics t () = write ~with_metrics !bindings_pool t
+
+  let write_seq ~with_metrics t =
+    let bindings = Array.copy !bindings_pool in
```
icristescu: I'm not sure about this. The point of `write_seq` is to see if an order on the keys has a better performance than `write_random`. But in  `write_seq` there are two (potentially big) arrays in the memory instead of one as in `write_random` and it affects the performance. 

Maybe sort the bindings_pool and have a flag which tells if you need to regenerate the array or not, in case `write_random` is called after a sort? Or order the benchmarks before calling them, so that the sorting benchmarks are called last?

bench/bench.ml
```diff
@@ -0,0 +1,484 @@
...
+        name = "replace_increasing_keys";
+        synopsis = "Replace in increasing order of keys";
+        readonly = false;
+        fresh = true;
+        benchmark = write_seq;
+        dependency = None;
+      };
+      {
+        name = "replace_increasing_hash";
+        synopsis = "Replace in increasing order of hash";
```
icristescu: "... order of hashes"

bench/bench.ml
```diff
@@ -0,0 +1,484 @@
...
+
+let init config =
+  Printexc.record_backtrace true;
+  Random.init config.seed;
+  if config.with_metrics then (
+    Metrics.enable_all ();
+    Metrics_gnuplot.set_reporter ();
+    Metrics_unix.monitor_gc 0.1 );
+  bindings_pool := make_bindings_pool config.nb_entries;
+  absent_bindings := make_bindings_pool config.nb_entries
```
icristescu: What do you think about generating this array only if the find absent benchmark is called? I think you can do that in the `run` function.

bench/bench.ml
```diff
@@ -0,0 +1,484 @@
...
+  let parse s =
+    try Ok Re.(compile @@ Pcre.re s) with
+    | Re.Perl.Parse_error -> Error (`Msg "Perl-compatible regexp parse error")
+    | Re.Perl.Not_supported -> Error (`Msg "unsupported regexp feature")
+  in
+  let print = Re.pp_re in
+  Arg.conv (parse, print)
+
+let name_filter =
+  let doc = "A regular expression matching the names of benchmarks to run" in
```
icristescu: It would be nice to have a way of only calling `replace_random` instead of calling both `replace_random` and `replace_random_sync`.

bench/bench.ml
```diff
@@ -0,0 +1,484 @@
...
+  let env = env_var "NB_ENTRIES" in
+  Arg.(value & opt int 10_000_000 & info [ "nb_entries" ] ~env ~doc)
+
+let list_cmd =
+  let doc = "List all available benchmarks." in
+  (Term.(pure list_benches $ const ()), Term.info "list" ~doc)
+
+let json_flag =
+  let doc = "Output the results as a json object." in
+  let env = env_var "JSON" in
```
icristescu: I think it would be easier to use the json output if it was written in a file. The option `-j` could ask for a filename, for instance?

bench/bench.ml
pascutto: You can use the end of string symbol `$`: `-f replace_random$`. (There is also `^` to force the beginning if you need.

CraigFe: Hmm, shouldn't `replace_random` work by itself? The regex that matches both of them is `replace_random.*`.

CraigFe: Hmm, shouldn't `replace_random` work by itself? The regex that matches both of them is `replace_random.*`.

CraigFe: Ah, the regex doesn't need to match the full name of the benchmark... I see.

icristescu: How do you use `^` exactly ? I can't find the right combination...

`$` is depend on the terminal, I have an error: `fish: Expected a variable name after this $.` (but maybe that is my problem :) ) 

Do you mind adding some usage examples?

bench/bench.ml
pascutto: I don't really like the idea of a `--json` flag - which sounds like a format flag -  that also kind of silently affects the output channel.
What would you think of a separate `-o`/`--output` option that lets you specify the output file? We could combine that with the `--json` flag to print the json output in a file.

icristescu: Yeah, that sounds good to me.

pascutto: Done :heavy_check_mark:

bench/bench.ml
pascutto: As discussed offline:
- We don't have to support terminal-specific issues like the one you experienced with `fish` (for the record, `$` had to be escaped), it's up to the user.
- I added a link to the Perl regexp syntax documentation

CraigFe: I think either:
 - the documentation should be updated to reflect that the regex only needs to match a subset of the bench name;
 - the filter should be changed to require a match on the entire name.

Currently, `--filter=""` runs every benchmark, which I find confusing.

bench/bench.ml
```diff
@@ -0,0 +1,506 @@
...
+  let env = env_var "NAME_FILTER" in
+  Arg.(
+    value
+    & opt (some regex) None
+    & info [ "f"; "filter" ] ~env ~doc ~docv:"NAME_REGEX")
+
+let data_dir =
+  let doc = "Set directory for the data files" in
+  let env = env_var "DATA_DIR" in
+  Arg.(value & opt dir "_bench" & info [ "d"; "data_dir" ] ~env ~doc)
```
CraigFe:
```suggestion
  Arg.(value & opt dir "_bench" & info [ "d"; "data-dir" ] ~env ~doc)
```

I think it's more conventional to use hyphens for linking words in these options (https://www.gnu.org/prep/standards/html_node/Option-Table.html#Option-Table).

bench/bench.ml
pascutto: Solved by using anchored regexps instead, so using `$` is not necessary anymore in your case @icristescu.

bench/bench.ml
pascutto: If we do that, we will face the same issue you described above: whether an absent bench runs or not will change the results of the other benches.

bench/bench.ml
pascutto: I solved this by having a third array in memory, dedicated to sorted benchs

PR #158, opened 2020-01-13
Make transitive dependencies explicit in dune files

CraigFe:
Having another go at this, using an explicit dependency on `stdlib-shims` to solve the "abstract `format4`" issue encountered last time (https://github.com/mirage/index/pull/142).


CraigFe:
Rebased.

PR #159, opened 2020-02-05
Remove file if it already exists.

gs0510:
The pipeline creates a file and then passes the path to the bench target to write the result to. It's better to have the control of creating the file in the pipeline since that can control file permissions, etc.


MagnusS:
Do we need a check for this at all? Wouldn't `open_out` just truncate the file if it exists, which would generally be what you'd expect would happen (similar to redirecting to an existing file)

pascutto:
It's probably still necessary to check that the given string is not an already existing directory, we would face an ugly error.

PR #161, opened 2020-02-06
Extract Thread functionality from the IO signature

CraigFe:
... and parameterise `Index.Make` over a thread implementation directly.

This is a natural extension of https://github.com/mirage/index/pull/160 (and is based on that PR for convenience). The threading implementation isn't strictly an 'IO' concern.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,8 +2,9 @@
 
 ## Changed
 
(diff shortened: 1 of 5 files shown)


CraigFe:
Rebased.

PR #164, opened 2020-02-19
Test (and document) when it is allowed to use the Make functor twice.

icristescu:


icristescu:
I'm closing this as there is a related [issue](https://github.com/mirage/irmin/issues/950) opened on irmin.

PR #168, opened 2020-03-02
Add after clear fails

icristescu:
If generation is `0`  and a clear is called, then the values added after the clear are not detected by an RO instance. 
 
A simple fix is to always reload the log when generation is `0`.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -459,7 +459,7 @@ struct
         let new_log_offset = IO.force_offset log.io in
         let add_log_entry e = add_log_entry log e in
         sync_log_async ~generation_change:(t.generation <> generation) ();
-        if t.generation <> generation then (
+        if t.generation <> generation || t.generation = 0L then (
           Log.debug (fun l ->
               l "[%s] generation has changed, reading log and index from disk"
                 (Filename.basename t.root));
--- a/src/index.ml
+++ b/src/index.ml
@@ -182,19 +182,22 @@ struct
     let t = check_open t in
     Log.debug (fun l -> l "clear %S" t.root);
     if t.config.readonly then raise RO_not_allowed;
-    t.generation <- 0L;
     Mutex.with_lock t.merge_lock (fun () ->
+        t.generation <- Int64.succ t.generation;
         let log = assert_and_get t.log in
-        IO.clear log.io;
+        IO.set_generation log.io t.generation;
+        IO.clear ~keep_generation:true log.io;
         Tbl.clear log.mem;
         may
           (fun l ->
-            IO.clear l.io;
+            IO.set_generation l.io t.generation;
+            IO.clear ~keep_generation:true l.io;
             IO.close l.io)
           t.log_async;
         may
           (fun (i : index) ->
-            IO.clear i.io;
+            IO.set_generation i.io t.generation;
+            IO.clear ~keep_generation:true i.io;
             IO.close i.io)
           t.index;
         t.index <- None;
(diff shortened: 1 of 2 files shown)


icristescu:
I’m proposing a second solution for this bug, which is that `clear`  increments  `generation` instead of setting it to `0`.  A generation changes means that a file was rewritten, and `clear` can be seen as that. It would also allow for a uniform use of generation in both index and irmin-pack, as proposed in [the clear stores PR](https://github.com/mirage/irmin/pull/939).
Let me know what you think, as it’s in a different commit, I can always revert back.

src/fan.mli
```diff
@@ -49,3 +49,6 @@ val import : hash_size:int -> string -> t
 (** [import ~hash_size buf] decodes [buf] such that
     [import ~hash_size (export t) = t] if [t] was initially created with
     ~hash_size. *)
+
+val clear : unit -> string
```
CraigFe, line 53:
```suggestion
val empty : string
```
I think this isn't clearing anything.

src/fan.ml
```diff
@@ -44,9 +44,11 @@ let nb_fans t = Array.length t.fans
 let fan t h = (h land t.mask) lsr t.shift
 
 let search t h =
-  let fan = fan t h in
-  let low = if fan = 0 then 0L else t.fans.(fan - 1) in
-  (low, t.fans.(fan))
+  if nb_fans t = 0 then (0L, 0L)
```
CraigFe, line 47: I think this upper bound is going to be interpreted inclusively by `interpolation_search`, so this is not correct.

src/fan.ml
```diff
@@ -104,5 +106,8 @@ let import ~hash_size buf =
   in
   let size = int_of_float (log2 (float_of_int nb_fans)) in
   let shift = hash_size - size in
-  let mask = (nb_fans - 1) lsl shift in
+  let nb_fans = max 0 (nb_fans - 1) in
+  let mask = nb_fans lsl shift in
```
CraigFe, line 110: I don't properly understand what's going on here, but I think `nb_fans` is a confusing name for `max 0 (nb_fans - 1)`.

src/index.ml
```diff
@@ -464,9 +468,10 @@ struct
           Tbl.clear log.mem;
           iter_io add_log_entry log.io;
           may (fun (i : index) -> IO.close i.io) t.index;
-          if Int64.equal generation 0L then t.index <- None
+          let index_path = index_path t.root in
+          if Int64.equal generation 0L || not (Sys.file_exists index_path) then
+            t.index <- None
```
CraigFe, line 473: Is this necessary now that the generation number is being incremented on clear?

src/unix/index_unix.ml
```diff
@@ -258,6 +258,7 @@ module IO : Index.IO = struct
 
   let clear ?(keep_generation = false) t =
     t.offset <- 0L;
+    t.fan_size <- 0L;
```
CraigFe, line 261: Is this an unrelated fix?

src/fan.mli
icristescu: You are right, and I actually don't need it.

src/fan.ml
icristescu: Good catch, thank you! I modified it to `(0L, 1L)`. (It's a bit annoying that no test failed because of this.)

src/fan.ml
icristescu: I wanted to avoid `nb_fans` and `mask` to become negative, but it seems that there are no repercussions on the tests either way so I removed it.

src/index.ml
icristescu: No, it seems not. (removed the test of generation)

src/fan.ml
CraigFe: Hmm, shouldn't it be `(0L, -1L)` if you want to make the range empty?

icristescu: > Good catch, thank you! I modified it to (0L, 1L). (It's a bit annoying that no test failed because of this.)

Sorry, made a bad manipulation :)

icristescu: > Hmm, shouldn't it be `(0L, -1L)` if you want to make the range empty?

Yes, indeed, but the failing test passed, I'll look more into it...

src/unix/index_unix.ml
icristescu: It's actually included in [`Fan.set`](https://github.com/mirage/index/blob/master/src/unix/index_unix.ml#L132), removed it as well.

src/fan.ml
icristescu: I couldn't reproduce the failing test for this (I think it is fixed by https://github.com/mirage/index/pull/168/files#diff-732d381615307f94d510e2dbdee74f4aR483) so I removed it.

PR #169, opened 2020-04-02
Timing replace operations

icristescu:
I’m using metrics to test this, so for
` dune exec -- ./bench/bench.exe -f replace_random_with_timer -m --nb-entries 100000`
I get, in the file `bench.data`
```
Timestamp ,_,_,_, nb replace, time of latest replace
6863370, 0, 0, 0, 1, 1.1e-05
6998576, 0, 0, 0, 2, 1e-06
7008271, 0, 0, 0, 3, 1e-06
7021192, 0, 0, 0, 4, 0.
7032426, 0, 0, 0, 5, 2.99999999999e-06
7043241, 0, 0, 0, 6, 0.
7052309, 0, 0, 0, 7, 2e-06
….
```
but it is very slow for a large number of entries!

### DIFF
--- a/bench/accuracy.ml
+++ b/bench/accuracy.ml
@@ -1,6 +1,0 @@
-let () =
-  let t1 = Sys.time () in
-  while t1 = Sys.time () do
-    ()
-  done;
-  Printf.printf "accuracy of sys time = %f" (Sys.time () -. t1)
--- a/bench/accuracy.ml
+++ b/bench/accuracy.ml
@@ -1,0 +1,6 @@
+let () =
+  let t1 = Sys.time () in
+  while t1 = Sys.time () do
+    ()
+  done;
+  Printf.printf "accuracy of sys time = %f" (Sys.time () -. t1)
--- a/bench/bench.ml
+++ b/bench/bench.ml
@@ -37,7 +37,7 @@ let random_char () = char_of_int (33 + Random.int 94)
 let random_string string_size =
   String.init string_size (fun _i -> random_char ())
 
-let replace_freq = ref 0
+let replace_sampling_interval = ref 0
 
 module Context = struct
   module Key = struct
@@ -157,10 +157,10 @@ module Index = struct
     let no_tags x = x in
     fun () -> Metrics.add src no_tags (fun m -> m (Stats.get ()))
 
-  let write ~with_metrics ?(with_flush = false) ?with_timer bindings rw =
+  let write ~with_metrics ?(with_flush = false) ?sampling_interval bindings rw =
     Array.iter
       (fun (k, v) ->
-        Index.replace_with_timer ?with_timer rw k v;
+        Index.replace_with_timer ?sampling_interval rw k v;
         if with_flush then Index.flush rw;
         if with_metrics then add_metrics ())
       bindings
@@ -182,7 +182,8 @@ module Index = struct
   let write_random ~with_metrics t () = write ~with_metrics !bindings_pool t
 
   let write_random_with_timer ~with_metrics t () =
-    write ~with_metrics ~with_timer:!replace_freq !bindings_pool t
+    write ~with_metrics ~sampling_interval:!replace_sampling_interval
+      !bindings_pool t
 
   let write_seq ~with_metrics t =
     Array.sort (fun a b -> String.compare (fst a) (fst b)) !sorted_bindings_pool;
@@ -351,7 +352,7 @@ type config = {
   log_size : int;
   seed : int;
   with_metrics : bool;
-  frequency : int;
+  sampling_interval : int;
 }
 [@@deriving yojson]
 
@@ -363,8 +364,8 @@ let pp_config fmt config =
      Log size: %d@\n\
      Seed: %d@\n\
      Metrics: %b@\n\
-     Frequency: %d" config.key_size config.value_size config.nb_entries
-    config.log_size config.seed config.with_metrics config.frequency
+     Sampling interval: %d" config.key_size config.value_size config.nb_entries
+    config.log_size config.seed config.with_metrics config.sampling_interval
 
 let cleanup root =
   let files = [ "data"; "log"; "lock"; "log_async"; "merge" ] in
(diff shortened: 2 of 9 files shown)


icristescu:
After discussing with @magnus, I’m adding a frequency for timing the replace operations. 
 
I added `accuracy.ml` so that you can check accuracy of `Sys.time` calls on your computer. For me its on average 2e-06s. If we time each replace operation, we often get 1e-06 or 0, which suggests `Sys.time` is not accurate enough for one operation. If instead we time replace operations by batches of 10 (or 100, 1000), one operation takes on average 1e-07, which seems to be the actual time of a replace, on average. 

I'm not sure we want to keep `accuracy.ml` or integrate it in some other form.

I’m adding the frequency as argument to `replace` so that we can adjust it during a run (for instance replaces seem to [take longer at the end of a run](https://gist.github.com/icristescu/973b1d523a7f1472c8d4305e5ba750b6), so we might want to have a smaller frequency at the end). 

You can see replace times either using metrics (slow but generates graphs) or by outputting to json.

samoht:
Can you try with [mtime](https://github.com/dbuenzli/mtime) instead of Sys.time? It has been designed for this use-case :-)

bench/accuracy.ml
```diff
@@ -0,0 +1,6 @@
+let () =
+  let t1 = Sys.time () in
+  while t1 = Sys.time () do
+    ()
+  done;
+  Printf.printf "accuracy of sys time = %f" (Sys.time () -. t1)
```
CraigFe, line 6: I think you're actually measuring precision here :slightly_smiling_face:

bench/bench.ml
```diff
@@ -491,6 +528,11 @@ let json_flag =
   let env = env_var "JSON" in
   Arg.(value & flag & info [ "j"; "json" ] ~env ~doc)
 
+let frequency =
+  let doc = "The frequency of timing the replace operations." in
```
CraigFe, line 532: 'frequency' to me implies dimensions of `time^{-1}` (i.e. bigger numbers correspond to more data points). Perhaps something like 'sampling interval' is a better term.

bench/bench.ml
```diff
@@ -491,6 +528,11 @@ let json_flag =
...
+  let env = env_var "FREQ" in
+  Arg.(value & opt int 10 & info [ "freq" ] ~env ~doc)
```
CraigFe, line 534: This flag (and its corresponding `env_var` seem a bit too generically named). Although `REPLACE_DURATION_SAMPLING_INTERVAL` might be too verbose :wink:

dune-project
```diff
@@ -2,3 +2,4 @@
 (name index)
 (using fmt 1.1)
 (implicit_transitive_deps false)
+(allow_approximate_merlin)
```
CraigFe, line 5: What's this for?

src/index.ml
```diff
@@ -708,8 +708,10 @@ struct
         Thread.return ()
     | Some witness -> merge ?hook ~witness t
 
-  let replace t key value =
+  let replace_with_timer ?with_timer t key value =
```
CraigFe, line 711: I feel that `replace_with_timer` should probably be implemented in terms of `replace` and not vice-versa. That way, this timing logic isn't touched in production and `?with_timer` doesn't need to be optional. I'm also suspicious of starting the timer after some of `replace` has already happened, in this case the `check_open`, and with the other implementation this can never occur.

CraigFe: Also, is it necessary to have the sampling logic inside Index at all? AFAICS, `replace` is only ever called externally. Perhaps it would be better to have `replace_timed : t -> k -> v -> <time interval>` and keep sampling considerations in the benchmarks.

bench/accuracy.ml
icristescu: Terminology is difficult :) I'm removing this file anyway, as we don't need it anymore with Mtime.

dune-project
icristescu: I had [this warning](https://discuss.ocaml.org/t/new-dune-warning-merlin-generated-is-inaccurate-pps-specification-isnt-identical-in-all-stanzas/3760) for which I don't know if there is another fix. Anyway, it does not matter anymore, since I remove the second executable (`accuracy`) from the `bench` directory, which was causing the problem.

src/index.ml
icristescu: > I feel that replace_with_timer should probably be implemented in terms of replace and not vice-versa. 

Yes, good point, thank you. 

> Also, is it necessary to have the sampling logic inside Index at all? AFAICS, replace is only ever called externally.

I'm not sure about this. It can be called from irmin, and the [initial motivation](https://ocamllabs.slack.com/archives/GNZ0111EH/p1585229720001200) for this was, I think to have these stats from tezos. I would like to have the sampling logic somewhere, to be used outside of our benchmarks.

icristescu:
> Can you try with [mtime](https://github.com/dbuenzli/mtime) instead of Sys.time? It has been designed for this use-case :-)

I didn't know about Mtime, thanks! 

I have to keep the `sampling interval` anyway. With our curent benchmarks we cannot store an additional list of `10_000_000` floats, it leads to `stack overflow`. So we have to compromise between benchmarking many operations with big sampling intervals, or benchmarking fewer operations with smaller sampling intervals.

src/index.ml
CraigFe: > I'm not sure about this. It can be called from irmin, and the initial motivation for this was, I think to have these stats from tezos. I would like to have the sampling logic somewhere, to be used outside of our benchmarks.

That's fair. In that case, I'd like some other name than `with_timer` (`sampling_interval`, perhaps?).

CraigFe:
> I have to keep the `sampling interval` anyway. With our curent benchmarks we cannot store an additional list of `10_000_000` floats, it leads to `stack overflow`. So we have to compromise between benchmarking many operations with big sampling intervals, or benchmarking fewer operations with smaller sampling intervals.

That sounds like a case of wanting to implement our own `List.foo` function somewhere. Regardless, we probably don't want the benchmarks to have a huge memory footprint anyway so I think the sampling is worthwhile :+1: Perhaps at some point in the future it'll be worthwhile to use [dynamic histograms](http://pages.cs.wisc.edu/~donjerko/hist.pdf) in our metrics.

icristescu:
> Perhaps at some point in the future it'll be worthwhile to use [dynamic histograms](http://pages.cs.wisc.edu/~donjerko/hist.pdf) in our metrics.

Maybe dynamic histograms are the way to go. Metrics also partially solve the issue, because data is written to disk at every add_metrics call, but it is slow.

PR #172, opened 2020-04-29
io_array tries to read from a possibly empty file

icristescu:
I'm running some [benchmarks for the layered store with an RO instance](https://github.com/icristescu/irmin/commit/2779ae93d22ce51fc512a69e2100a5cb276f05eb) and I get 
```
+1588092108us index [DEBUG] No existing buffer. Prefetching in range [0, 0]
Benchmarks for layered store: internal error, uncaught exception:
                              File "src/io_array.ml", line 89, characters 4-10: Assertion failed
```
and after adding more logs it seems that is the interpolation search that sets `low` and `high` to 0
```
+1588163729us index [DEBUG] No existing buffer. Prefetching in range [0, 0]
Benchmarks for layered store: internal error, uncaught exception:
                              Index__Io_array.Make(IO)(Elt).Io_Array("search key low 0 high 0")
```
I couldn't reproduce this with the concurrent tests in index, but it seems anyway an error that `range` is not 0 when both `low` and `high` are.

### DIFF
--- a/src/io_array.ml
+++ b/src/io_array.ml
@@ -85,9 +85,11 @@ module Make (IO : Io.S) (Elt : ELT) :
     let range = Elt.encoded_size * (1 + Int64.to_int (high -- low)) in
     let low_off = Int64.mul low Elt.encoded_sizeL in
     let high_off = Int64.mul high Elt.encoded_sizeL in
-    let n = IO.read t.io ~off:low_off ~len:range buf in
-    assert (n = range);
-    t.buffer <- Some { buf; low_off; high_off }
+    if high -- low = Int64.zero then ()
+    else
+      let n = IO.read t.io ~off:low_off ~len:range buf in
+      assert (n = range);
+      t.buffer <- Some { buf; low_off; high_off }
 
   let pre_fetch t ~low ~high =
     let range = Elt.encoded_size * (1 + Int64.to_int (high -- low)) in


CraigFe:
> I couldn't reproduce this with the concurrent tests in index, but it seems anyway an error that `range` is not 0 when both `low` and `high` are.

`low` and `high` are being interpreted interpreted elsewhere are two inclusive bounds, so a range `[x, x]` contains the single element at index `x` (c.f. https://github.com/mirage/index/blob/master/src/search.ml#L124-L126) – this doesn't seem to be documented anywhere.

Perhaps `io_array` should guard more carefully against the element being searched for not actually existing – again, the [documentation](https://github.com/mirage/index/blob/master/src/search.mli#L67-L68) has nothing to say about this.

pascutto:
I think we should maintain the invariant that `t.index` should be `None` when the index is empty. We are already doing this in _most_ cases, but forgot to check that in the case the index starts and an empty index file already exists (this should set the index to `None`) [here](https://github.com/mirage/index/blob/master/src/index.ml#L379-L392). 
I think the correct fix is to have `index` to `None` if `entries` is `0` (L383).

Does this happen in such a configuration (eg clear and reopen)?

As a side note this whole mechanism won't be needed anymore if we get rid of the multiple instances support.

pascutto:
e.g. this test:
```ocaml
  let _clear_close_reopen () =
    let Context.{ clone; rw; _ } = Context.full_index () in
    Index.clear rw;
    Index.close rw;
    let rw2 = clone ~fresh:false ~readonly:false () in
    Alcotest.check_raises "Finding absent should raise Not_found" Not_found
      (fun () -> Key.v () |> Index.find rw2 |> ignore_value);
    Index.close rw2
```
fails, but with `array out of bounds`, so it's maybe not exactly the same issue?

CraigFe:
> fails, but with `array out of bounds`, so it's maybe not exactly the same issue?

I suspect it's a subset in which the absent value is outside the range of the array entirely. This test doesn't seem to be covering that case.

pascutto:
The latter is fixed by setting `index` to `None` if it's empty. @icristescu could you check if it also fixes your use case?

icristescu:
I'm closing this as I can't reproduce the error and some of the fixes proposed here were already added by PRs https://github.com/mirage/index/pull/183 and https://github.com/mirage/index/pull/168

PR #173, opened 2020-05-13
Move IO implementation to index

pascutto:


CraigFe:
Looking back at this PR, I suspect the `Provider` interface is too Unix-specific to permit in-memory and Mirage-compatible implementations. `Irmin_unix.IO` is perhaps best kept where it is for now.

PR #175, opened 2020-06-12
RO sync

icristescu:
An RO syncs  when : (1) opening the store and (2) when calling `ro_syncs`.

There are some tests that fail due to `clear`, so I commented them out (to make them work, we should rebase this over https://github.com/mirage/index/pull/168 ). But if you're not calling clear, this should work.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -219,6 +219,80 @@ struct
           Int64.of_float rounded
       end)
 
+  let try_load_log t path =
+    Log.debug (fun l ->
+        l "[%s] checking on-disk %s file" (Filename.basename t.root)
+          (Filename.basename path));
+    if Sys.file_exists path then (
+      let io =
+        IO.v ~fresh:false ~readonly:true ~generation:0L ~fan_size:0L path
+      in
+      let mem = Tbl.create 0 in
+      iter_io (fun e -> Tbl.replace mem e.key e.value) io;
+      Some { io; mem } )
+    else None
+
+  let sync_log t =
+    Log.debug (fun l ->
+        l "[%s] checking for changes on disk" (Filename.basename t.root));
+    let no_changes () =
+      Log.debug (fun l ->
+          l "[%s] no changes detected" (Filename.basename t.root))
+    in
+    let add_log_entry log e = Tbl.replace log.mem e.key e.value in
+    let sync_log_async ?(generation_change = false) () =
+      match t.log_async with
+      | None -> t.log_async <- try_load_log t (log_async_path t.root)
+      | Some log ->
+          let offset = IO.offset log.io in
+          let new_offset = IO.force_offset log.io in
+          if generation_change || offset <> new_offset then (
+            Tbl.clear log.mem;
+            iter_io (add_log_entry log) log.io )
+          else ()
+    in
+    ( match t.log with
+    | None -> t.log <- try_load_log t (log_path t.root)
+    | Some _ -> () );
+    match t.log with
+    | None -> sync_log_async ()
+    | Some log ->
+        let generation = IO.get_generation log.io in
+        let log_offset = IO.offset log.io in
+        let new_log_offset = IO.force_offset log.io in
+        let add_log_entry e = add_log_entry log e in
+        sync_log_async ~generation_change:(t.generation <> generation) ();
+        if t.generation <> generation then (
+          Log.debug (fun l ->
+              l "[%s] generation has changed, reading log and index from disk"
+                (Filename.basename t.root));
+          t.generation <- generation;
+          Tbl.clear log.mem;
+          iter_io add_log_entry log.io;
+          may (fun (i : index) -> IO.close i.io) t.index;
+          let index_path = index_path t.root in
+          if not (Sys.file_exists index_path) then t.index <- None
+          else
+            let io =
+              IO.v ~fresh:false ~readonly:true ~generation ~fan_size:0L
+                index_path
+            in
+            let fan_out =
+              Fan.import ~hash_size:K.hash_size (IO.get_fanout io)
+            in
+            if IO.offset io = 0L then t.index <- None
+            else t.index <- Some { fan_out; io } )
+        else if log_offset < new_log_offset then (
+          Log.debug (fun l ->
+              l "[%s] new entries detected, reading log from disk"
+                (Filename.basename t.root));
+          iter_io add_log_entry log.io ~min:log_offset )
+        else if log_offset > new_log_offset then
+          (* In that case the log has probably been emptied and is being
+             refilled with async_log contents. *)
+          no_changes ()
+        else no_changes ()
+
   let with_cache ~v ~clear =
     let roots = Hashtbl.create 0 in
     let f ?auto_flush_callback ?(fresh = false) ?(readonly = false) ~log_size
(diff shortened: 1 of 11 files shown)


src/index.ml
```diff
@@ -296,6 +298,83 @@ struct
...
+  let try_load_log t path =
+    Log.debug (fun l ->
+        l "[%s] checking on-disk %s file" (Filename.basename t.root)
+          (Filename.basename path));
+    if Sys.file_exists path then (
+      let io =
+        IO.v ~fresh:false ~readonly:true ~generation:0L ~fan_size:0L path
+      in
+      let mem = Tbl.create 0 in
+      iter_io (fun e -> Tbl.replace mem e.key e.value) io;
```
samoht, line 310: Why do you need `replace` here?  Would `add` also work? If yes we should be using `add` as it is much more efficient.

src/index.ml
```diff
@@ -296,6 +298,83 @@ struct
...
+    else None
+
+  let sync_log t =
+    Log.debug (fun l ->
+        l "[%s] checking for changes on disk" (Filename.basename t.root));
+    let no_changes () =
+      Log.debug (fun l ->
+          l "[%s] no changes detected" (Filename.basename t.root))
+    in
+    let add_log_entry log e = Tbl.replace log.mem e.key e.value in
```
samoht, line 321: Same question with `replace` vs. `add`

src/index.ml
```diff
@@ -782,6 +783,18 @@ struct
                 t.log;
               may (fun (i : index) -> IO.close i.io) t.index;
               may (fun lock -> IO.unlock lock) t.writer_lock ))
+
+  let ro_sync t =
+    Stats.incr_nb_ro_sync ();
+    let t = check_open t in
+    Log.info (fun l -> l "[%s] ro_sync" (Filename.basename t.root));
+    if t.config.readonly then sync_log t
+    else Fmt.failwith "only an RO instance should call this function"
```
samoht, line 792: `Fmt.invalid_arg` is probably better here to indicate a user error (e.g. a wrong use of the API, not a bug in our code).

src/index.ml
icristescu: I changed `replace` with `add`, but there is still one place where we still want to use `replace` (https://github.com/mirage/index/pull/175/files#diff-732d381615307f94d510e2dbdee74f4aR371). When we add new values, we want to replace old binding with new ones.

samoht: What's the semantics that we want? yes, `Tbl.add` will keep multiple entries for the keys, but `Tbl.find` will always return the most recent one, which is exactly what we want I think. We just need to be careful with duplicated entries if we do a `Tbl.iter` later on. I don't know if it's a good idea or not, but `replace` can be very costly when the hashtbl buckets starts to be full. (probably something to benchmarks to see if it's worth it)

samoht: Note: I'm fine to do these changes in a separate PR, as it has nothing todo with the explicit `ro_sync`

icristescu: I think `iter` is the only place (in the RO instance) where we care about duplicate entries, but we say that we do not guarantee this https://github.com/mirage/index/blob/master/src/index.mli#L176

I think we should have this changes in a different PR as the RW instance also uses `Tbl.replace` in a few places. I'll revert these changes in this PR.

pascutto: Please bear in mind that whenever a `merge` happens, this table folded into an array, sorted, and used to fill the new index file. At that point if you have multiple bindings for a key they will all be folded and written to disk during the merge, and you can retrieve them by reading the index, which is not what we want.

icristescu: > I'll revert these changes in this PR.

Done. (I'll open a replace/add PR).

src/index.ml
```diff
@@ -323,6 +403,7 @@ struct
       with Not_found ->
         let instance = v ~fresh ~readonly ~log_size root in
         Hashtbl.add roots (root, readonly) instance;
+        if readonly then sync_log instance;
```
CraigFe, line 406: Bit difficult to see what's going on here, but doesn't this try to sync the log after creating a fresh instance on disk? (In at least two code paths, but maybe not all.)

src/stats.ml
```diff
@@ -67,3 +75,12 @@ let end_replace ~sampling_interval =
...
     nb_replace := 0 )
+
+let ro_sync_timer = ref (Mtime_clock.counter ())
+
+let start_ro_sync () = ro_sync_timer := Mtime_clock.counter ()
+
+let end_ro_sync () =
+  let span = Mtime_clock.count !ro_sync_timer in
+  let time = Mtime.Span.to_us span in
+  stats.ro_sync_times <- time :: stats.ro_sync_times
```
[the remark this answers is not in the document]
icristescu: Thanks, much cleaner.

src/index.mli
```diff
@@ -188,6 +188,9 @@ module type S = sig
 
   val close : t -> unit
   (** Closes all resources used by [t]. *)
+
+  val ro_sync : t -> unit
+  (** [ro_sync t] syncs an RO instance with the files on disk. *)
```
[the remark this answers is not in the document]
icristescu: I agree so I removed it.

src/index.ml
icristescu: It can be that the rw instance is already open. When the ro instance is opened, it does not automatically sync its log with the one on disk (see https://github.com/mirage/index/blob/master/src/index.ml#L340). 
 
I added a [test](https://github.com/mirage/index/pull/175/files#diff-22a049f4c6a8354008c166e8f65b02b6R268) that fails without this.

test/unix/main.ml
```diff
@@ -257,18 +259,99 @@ module Readonly = struct
...
+    Index.replace rw k1 v1;
+    Index.flush rw;
+    Index.ro_sync ro;
+    check_index_entry ro k1 v1;
+    Index.clear rw;
+    let k2, v2 = (Key.v (), Value.v ()) in
+    Index.replace rw k2 v2;
+    Index.flush rw;
+    Index.ro_sync ro;
+    check_index_entry ro k2 v2
```
pascutto, line 328: Missing `close`s here.
And perhaps that could check that `k1, v1` can't be found anymore?

src/index.ml
```diff
@@ -506,16 +514,10 @@ struct
...
         find_if_exists ~name:"log_async"
           ~find:(fun log -> Tbl.find log.mem)
           t.log_async
-        @~ fun () ->
-        find_log_index @~ fun () ->
-        if t.config.readonly then (
-          sync_log t;
-          find_log_index () )
-        else raise Not_found)
+        @~ find_log_index)
```
pascutto, line 520: The case that was handled by this retry is now completely omitted.
That means that it is virtually possible to call `ro_sync` and have the same interleaving behavior with `merge` as before, causing the `sync_log` to not see the disk modifications (that's why we retried before).

src/index.ml
```diff
@@ -782,6 +783,15 @@ struct
                 t.log;
               may (fun (i : index) -> IO.close i.io) t.index;
               may (fun lock -> IO.unlock lock) t.writer_lock ))
+
+  let ro_sync t =
+    Stats.incr_nb_ro_sync ();
+    let t = check_open t in
+    Log.info (fun l -> l "[%s] ro_sync" (Filename.basename t.root));
+    if t.config.readonly then sync_log t
+    else invalid_arg "only an RO instance should call this function"
```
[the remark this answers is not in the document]
icristescu: Replaced this with a `RW_not_allowed` exception.

src/index.ml
icristescu: Yes, and it's tricky, because no test fails (I run them 500 times). 
 
I looked at what interleaving execution was causing this bug, 
(https://github.com/mirage/index/pull/118#issuecomment-548083904), and this particular bug can be fixed at the end of a merge, by first setting the generation number and then clearing the log. It can cause a ro_sync to refill a log, even if there are no new values, but it cannot lead to a value not found anymore. What do you think?

CraigFe: As discussed offline: we can test this properly with a couple of extra hooks, and we probably should.

src/index.mli
```diff
@@ -178,16 +182,18 @@ module type S = sig
...
   val flush : ?with_fsync:bool -> t -> unit
   (** Flushes all internal buffers of the [IO] instances. If [with_fsync] is
       [true], this also flushes the OS caches for each [IO] instance. *)
 
   val close : t -> unit
   (** Closes all resources used by [t]. *)
+
+  val ro_sync : t -> unit
+  (** [ro_sync t] syncs a read-only index with the files on disk. Raises
+      [RW_not_allowed] if called by an read-write index. *)
```
CraigFe, line 196:
```suggestion
      [RW_not_allowed] if called by a read-write index. *)
```

src/index.ml
icristescu: I added the hooks and tested them in commit https://github.com/mirage/index/pull/175/commits/8d0f15bc1ac8ad1ebf8313f00b85e811e7e22168. 
There is one test that fails in that commit: the RO reads the generation number and the offset to decide if a change occurred. If between the two reads both the generation changed and the log was cleared then RO cannot detect a change (see test https://github.com/mirage/index/pull/175/commits/8d0f15bc1ac8ad1ebf8313f00b85e811e7e22168#diff-81aec069a46f58216cc2e7046f8d183dR349). 
The solution I propose in https://github.com/mirage/index/pull/175/commits/0af8704f7d3f3626f1d99ff883d3373d9fd718c6 is to group generation and offset reads/writes into a single read/write.

icristescu: I knew you had this somewhere, but didn't know where :) I added some of your modifications in the commit, and added you as co-author. 
I did not add the `fan_size` in the headers, as it is not necessary for the correctness of this PR.

src/index.ml
```diff
@@ -736,6 +735,16 @@ struct
...
 
   let close = close' ~hook:(fun _ -> ())
+
+  let sync' ?hook t =
+    Stats.incr_nb_sync ();
+    let t = check_open t in
+    Log.info (fun l -> l "[%s] sync" (Filename.basename t.root));
+    if t.config.readonly then sync_log ?hook t else raise RW_not_allowed
+
+  let sync t = sync' t
```
samoht, line 745: the `sync'` function doesn't seem necessary

src/index_intf.ml
```diff
@@ -268,6 +278,13 @@ module type Index = sig
           of consecutive operations, which can be specified by
           [sampling_interval]. If [sampling_interval] is not set, no operation
           is timed. *)
+
+      val sync_with_timer : t -> unit
+      (** Time sync operations. *)
+
+      val sync' : ?hook:[ `Before_offset_read ] Hook.t -> t -> unit
```
samoht, line 285: can you expose `sync` instead of `sync`` ? Also do why do we need a `with_timer` and a non-`with-timer` one? Maybe always using the timers is a good idea?

src/stats.ml
```diff
@@ -67,3 +75,10 @@ let end_replace ~sampling_interval =
     stats.replace_times <- average :: stats.replace_times;
     replace_timer := Mtime_clock.counter ();
     nb_replace := 0 )
+
+let sync_with_timer f =
+  let timer = Mtime_clock.counter () in
+  f ();
+  let span = Mtime_clock.count timer in
+  let time = Mtime.Span.to_us span in
+  stats.sync_times <- time :: stats.sync_times
```
samoht, line 84: This looks like an unbounded list and should probably be documented somewhere

samoht:
That looks good to me and the tests are all green, so feel free to merge.

src/index.ml
icristescu: `sync` is hiding the hook used in `sync'` in the public API

src/stats.ml
icristescu: Replaced this with a timer for only the latest sync call (instead of storing all previous sync calls). Also, by default every call to sync is timed now, so I removed the `sync_with_timer`.

src/index.ml
```diff
@@ -298,7 +375,7 @@ struct
                 append_key_value log.io e.key e.value)
               io;
             IO.sync log.io;
-            IO.clear io)
+            IO.clear ~generation:0L io)
```
CraigFe, line 378: I believe you're correctly preserving the existing semantics, but why don't we increment the generation number here?

icristescu: We are only increasing the generation on clear when calling `clear` in index (as opposed to whenever we call `IO.clear`). 
I changed this line with the last commit, so that `log_async` uses the same generation number as index and log. But it doesn't change too much, we never read the generation from the log_async.

CraigFe: OK :+1: Thanks for the explanation. This will make more sense in the code when we have better distinction between IO and header formats.

CraigFe:
Will merge once Travis passes.

CraigFe:
Thanks!

PR #176, opened 2020-06-12
Expose syscall bindings in the public API

CraigFe:
This is a step towards: https://github.com/mirage/irmin/issues/1009.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
 - Added `filter`, removing bindings depending on a predicate (#165)
 
+- Added `Irmin_unix.Syscalls`, a module exposing various Unix bindings for
+  interacting with file-systems.
+
 ## Changed
 
 - Parameterise `Index.Make` over arbitrary mutex and thread implementations
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,11 +1,15 @@
+# Unreleased
+
+## Added
+
+- Added `Index_unix.Syscalls`, a module exposing various Unix bindings for
+  interacting with file-systems.
+
 # 1.2.0 (2020-02-25)
 
 ## Added
 
 - Added `filter`, removing bindings depending on a predicate (#165)
-
-- Added `Index_unix.Syscalls`, a module exposing various Unix bindings for
-  interacting with file-systems.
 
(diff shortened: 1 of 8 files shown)


CHANGES.md
```diff
@@ -4,6 +4,9 @@
 
 - Added `filter`, removing bindings depending on a predicate (#165)
 
+- Added `Irmin_unix.Syscalls`, a module exposing various Unix bindings for
```
samoht, line 7: Index_unix I guess?

CraigFe: Ah, indeed. Fixed.

CraigFe:
Will merge this once the CI passes, unless there are any objections.

src/unix/syscalls.ml
```diff
@@ -0,0 +1,13 @@
...
+let pread ~fd ~fd_offset ~buffer ~buffer_offset ~length =
+  pread fd fd_offset buffer buffer_offset length
+
+external pwrite : Unix.file_descr -> int64 -> bytes -> int -> int -> int
+  = "caml_pwrite"
+
+let pwrite ~fd ~fd_offset ~buffer ~buffer_offset ~length =
+  pwrite fd fd_offset buffer buffer_offset length
+
+external fsync : Unix.file_descr -> unit = "unix_fsync"
```
samoht, line 13: as we are exporting these symbols a bit more publicly now, maybe it's a good time to check that have consistent names. Why `caml_pread` but `unix_fsync`? :-)

samoht: we should make sure they won't clash with symbols defined in `tezos-stdlib-unix` and `extunix`

CraigFe: `tezos-stdlib-unix` doesn't have any external definitions. `extunix` has the following:
- `caml_extunix_fsync`
- `caml_extunix_pread64`
- `caml_extunix_pwrite64`

I have no particular thoughts on what sort of naming scheme we should use. Anyone want to give an opinion?

samoht: use `caml_index_{op}` seems the safest bet.

CraigFe: Done.

PR #177, opened 2020-06-12
Extract Raw module

CraigFe:
This extracts the `Raw` module to a `raw.ml` file, and exposes it as
`Private.Raw`. This module is now identical to the one in `Irmin_pack`, so we
might as well pull it from here to stop them from diverging again in future.

Depends on https://github.com/mirage/index/pull/176.

### DIFF
--- a/src/unix/raw.mli
+++ b/src/unix/raw.mli
@@ -1,0 +1,51 @@
+(** [Raw] wraps a file-descriptor with an file-format used internally by Index.
+    The format contains the following header fields:
+
+    - {b offset}: a 64-bit integer, denoting the length of the file containing
+      valid data;
+    - {b version}: an 8-byte version string;
+    - {b generation}: a 64-bit integer denoting the generation number;
+    - {b fan}: a 64-bit length field, followed by a string containing that many
+      bytes. *)
+
+type t
+(** The type of [raw] file handles. *)
+
+val v : Unix.file_descr -> t
+(** Construct a [raw] value from a file descriptor. *)
+
+val unsafe_write : t -> off:int64 -> string -> unit
+
+val unsafe_read : t -> off:int64 -> len:int -> bytes -> int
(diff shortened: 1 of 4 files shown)


CraigFe:
As pointed out by @icristescu in https://github.com/mirage/irmin/pull/1010, the `cursor` record field in this module is actually unused. This PR now drops that field.

src/unix/raw.mli
```diff
@@ -0,0 +1,51 @@
...
+
+val close : t -> unit
+
+module Version : sig
+  val get : t -> string
+
+  val set : t -> string -> unit
+end
+
+module Offset : sig
```
samoht, line 31: That API is not great, as it forces to use 3 short reads for data which live on the same page - could we have a `Header` module instead, which reads these 3 fields in one go? (need to check exactly how it's used in index and try to minimise the number of short reads)

CraigFe: I've pushed a new commit that adds a new method for reading all of the fields in a single pass. My view is that this ends up not being worth it. It saves one read at the end of a merge and ~3 reads when constructing an instance; neither of which is particularly frequent.

The implementation ends up needing to do some annoying book-keeping to hide the `fan_size` header from the user.

samoht: The only place where this is important is https://github.com/mirage/index/pull/177/commits/6fe96bd38cae5e16981367c47d93e24a36a715b8#diff-732d381615307f94d510e2dbdee74f4aR459-R461 as `sync` is currently called on every reads for read-only instances. So we don't really care about fans. Also, with https://github.com/mirage/index/pull/175 maybe we don't care at all about optimising operations on these headers. So can you revert your last commit? We will see after #175 if something like this is still needed. Thanks!

CraigFe: Done.

PR #178, opened 2020-06-12
Fail when Unix IO version number is not as expected

CraigFe:
It's now no longer necessary to keep the version number as an IO field,
since it will always be equal to `current_version`.


samoht:
Failing this early means that the code which deals with version upgrade will have to be called quite early in the initialisation process -- but that code doesn't exist yet, so we will see when that happen.

PR #179, opened 2020-06-12
Add missing mtime dependency to opam file

CraigFe:
PR https://github.com/mirage/index/pull/169 added a dependency on `mtime` which is not tracked in the `opam` file. (That was actually picked up by the CI at the time, but we missed it...)


CraigFe:
(Note that this wasn't picked up by OCaml-CI since it doesn't isolate packages (and `index-unix.opam` already depends on `mtime` via `metrics-unix`).

PR #181, opened 2020-06-16
Tbl.add instead of Tbl.replace for RO instances

icristescu:
We can use `Tbl.add` instead of `Tbl.replace` when we are adding values from a log or index (where there are no duplicate values) into a clear table. 

In the RO instance (in `sync_log`) we can always use `Tbl.add`.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -218,12 +218,14 @@ struct
         IO.v ~fresh:false ~readonly:true ~generation:0L ~fan_size:0L path
       in
       let mem = Tbl.create 0 in
-      iter_io (fun e -> Tbl.replace mem e.key e.value) io;
+      (* We use Tbl.add only in RO instances (here and in [sync_log]), as the RW
+         has to ensure that there are no duplicates in the log. *)
+      iter_io (fun e -> Tbl.add mem e.key e.value) io;
       Some { io; mem })
     else None
 
   let sync_log_entries ?min log =
-    let add_log_entry (e : Entry.t) = Tbl.replace log.mem e.key e.value in
+    let add_log_entry (e : Entry.t) = Tbl.add log.mem e.key e.value in
     if min = None then Tbl.clear log.mem;
     iter_io ?min add_log_entry log.io
 
@@ -790,7 +792,6 @@ struct
     | None -> ()
     | Some log ->
         Tbl.iter f log.mem;
-        may (fun (i : index) -> iter_io (fun e -> f e.key e.value) i.io) t.index;
         Semaphore.with_acquire t.rename_lock (fun () ->
             (match t.log_async with
             | None -> ()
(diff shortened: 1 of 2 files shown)


src/index.ml
```diff
@@ -345,7 +345,7 @@ struct
             l "[%s] log file detected. Loading %Ld entries"
               (Filename.basename root) entries);
         let mem = Tbl.create (Int64.to_int entries) in
-        iter_io (fun e -> Tbl.replace mem e.key e.value) io;
+        iter_io (fun e -> Tbl.add mem e.key e.value) io;
```
pascutto, line 348: This is RW code so it will ultimately lead to duplicates in the index file.

icristescu: But it is loading entries from the log file, which shouldn't contain duplicates, no?

icristescu: Yes, you are right. Thanks !

I changed it back and also added two tests for duplicate entries, (one that fails if add is used here). In the test I noticed that `iter` calls iter on index twice, so I remove one of them.

src/index.ml
```diff
@@ -753,7 +755,6 @@ struct
     | None -> ()
     | Some log ->
         Tbl.iter f log.mem;
-        may (fun (i : index) -> iter_io (fun e -> f e.key e.value) i.io) t.index;
```
samoht, line 756: why did you remove this line?

icristescu: the iter on index is called again a bit lower (https://github.com/mirage/index/blob/master/src/index.ml#L761), but git diff is not showing it.

samoht: yes but why is this necessary to remove that line? Is that related to the change between replace and add?

icristescu: I think these tests are related to this PR, but I can move both the tests and the change to `iter` in a different PR, if you think is clearer.

samoht: well that's a bit suspicious if we have to change the semantics of a function to make the tests pass :-) @pascutto do you remember why `iter` is iterating over the elements twice? That sounds like a bug?

pascutto: I'll check what's going on here.
A side note though: `iter` may hit some keys twice even if there are in fact no duplicates so it's not super robust to write tests assuming that `iter` reflects the correct state of duplication in the structure. (see our docstring here: https://github.com/mirage/index/blob/master/src/index.mli#L176-L183)

samoht:
I still interested to run benchmarks with that PR if you case you have a bit of time to rebase it :-)

icristescu:
I rebased this, but it only replaces `Tbl.replace` for RO instances (updated the title of the PR to reflect that), so it's probably not going to change much in the layers benchmark.

samoht:
Closing this one as it only applies to read-only instance (and it's not clear it super useful).

PR #182, opened 2020-06-18
Clear + RO

icristescu:
This PR uses https://github.com/mirage/index/pull/175 and https://github.com/mirage/index/pull/168 to make sure we have tested the two together (the tests are in `test/unix/main.ml`)

We are testing that 
1. after a `clear` and an `ro_sync` the latest values are found and old values are not found;
2. after a `clear` but *before* a `ro_sync`, old values are still found by the RO. 

Scenario 2 seems to hold because either old values are in `log.mem` (which is not updated unless an explicit call to `ro_sync`)  or they are in index. But the index is only updated after a merge, which renames the merge file into index. Meaning that the RO still reads the old values from the previous version of index, until it explicitly calls `ro_sync`. 

Did I miss some cases ?

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -101,6 +101,8 @@ module type S = sig
   val flush : ?with_fsync:bool -> t -> unit
 
   val close : t -> unit
+
+  val ro_sync : t -> unit
 end
 
 let may f = function None -> () | Some bf -> f bf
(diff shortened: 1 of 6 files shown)


icristescu:
These tests are now included in https://github.com/mirage/index/pull/175.

Issue #180, opened 2020-06-15
Allow Index.close to abort ongoing merges


CraigFe:
Currently the `close : t -> unit` operation blocks until any ongoing merge operations have completed (by trying to seize the merge lock). This isn't necessary for integrity of the store: a partial merge could be terminated as part of the `close` operation, without corrupting any of the data.

We should support something like an `?abort` option to `close` that causes it to exit as quickly as possible without corrupting the store:
```ocaml
val close : ?abort:bool -> t -> unit
```
This would be useful when `close`-ing as part of a `SIGTERM` handler.

PR #185, opened 2020-06-25
Allow merge operations to yield to a concurrent close

CraigFe:
Fix https://github.com/mirage/index/issues/180.

The change is in two parts:

1. generalise threads to return results from children;

1. implement a pre-emptable merge operation (using thread results to propagate the return status of asynchronous merges back to the parent thread).

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -62,13 +62,13 @@ module type MUTEX = sig
 end
 
 module type THREAD = sig
-  type t
+  type 'a t
 
-  val async : (unit -> 'a) -> t
+  val async : (unit -> 'a) -> 'a t
 
-  val await : t -> unit
+  val await : 'a t -> ('a, [ `Async_exn of exn ]) result
 
-  val return : unit -> t
+  val return : 'a -> 'a t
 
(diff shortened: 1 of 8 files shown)


icristescu:
Maybe it is a stupid question, but why not use `thread.kill`? We can keep the thread that is doing the merge as mutable field in index and kill it in the close operation, no?

CraigFe:
Not a stupid question at all. I considered it, but there are a few reasons for not going with that approach:

- We need to be sure the merge lock is treated properly. `Thread.kill` might kill the child while it's still holding the merge lock, without a good way to determine when that's happened.

- `Thread.kill` can't distinguish an already-terminated thread, making it impossible to return ``[ `Completed | `Aborted ]``. That's not a huge problem, but is harder to test.

- (bikeshedding) For blocking merges, there is no child thread to kill. It's technically possible for that to be a problem when using hooks.

- (bikeshedding) The Lwt equivalent of `Thread.kill` is `Lwt.cancel`, which has the same ergonomics problems and also doesn't always work (threads can be made 'uncancelable').

PR #186, opened 2020-06-25
De-duplicate signatues in {index,search}.mli

CraigFe:
Pure refactor; no behavioural changes. Create some `_intf` files to avoid duplicating large module types in `.{ml,mli}` pairs. We need to do some gymnastics to avoid linking to those files in the docs (but once https://github.com/ocaml/odoc/pull/439 is released we won't need to do even that).

Depends on https://github.com/mirage/index/pull/185, for no particular reason other than my own laziness...


src/search.ml
```diff
@@ -15,70 +15,18 @@ furnished to do so, subject to the following conditions:
...
-end
+include Search_intf
 
 module Make
     (Entry : ENTRY)
     (Array : ARRAY with type elt = Entry.t)
     (Metric : METRIC with module Entry := Entry) :
   S with module Entry := Entry and module Array := Array = struct
-  module Entry = Entry
-  module Array = Array
```
CraigFe, line 81: Tiny code change here. We don't need to create these modules in the `struct` since we used destructive substitution in the return type.

src/index_intf.ml
```diff
@@ -0,0 +1,271 @@
...
+module type Index = sig
+  (** The input of [Make] for keys. *)
+  module type Key = sig
+    (* N.B. We use [sig ... end] redirections to avoid linking to the [_intf]
+       file in the generated docs. Once Odoc 2 is released, this can be
+       removed. *)
+
+    include Key
+    (** @inline *)
+  end
```
CraigFe, line 175: Here's the gymnastics in question. It's sufficient to do:

```ocaml
module type Key = Key
```

but then Odoc will add an extra link indirection through the `_intf` file: not a great UX.

PR #188, opened 2020-06-26
Handle sharing of Index instances explicitly

CraigFe:
Add a new `Cache.S` signature with two implementations: `Noop` and `Unbounded`. A cache implementation must now be provided to `Index.Make`.

The explicit instance sharing is Implemented with an _optional_ `?cache` argument to `Index.v`, from which the instance may be picked (and to which any new instances will be added). `cache` is perhaps not the best name for this argument: I picked it just because that's the necessary semantics for the underlying datastructure. (`pool`? `sharing_pool`? Ideas welcome.)

### DIFF
--- a/src/index_intf.ml
+++ b/src/index_intf.ml
@@ -134,7 +134,7 @@ module type S = sig
   (** The constructor for indexes.
 
       @param auto_flush_callback adds a callback before an auto flush.
-      @param cache a cache instance to use for instance sharing.
+      @param cache used for instance sharing.
       @param fresh whether an existing index should be overwritten.
       @param read_only whether read-only mode is enabled for this index.
       @param log_size the maximum number of bindings in the `log` IO. *)
--- a/src/index_intf.ml
+++ b/src/index_intf.ml
@@ -117,8 +117,15 @@ module type S = sig
   type value
   (** The type for values. *)
 
+  type cache
+  (** The type for caches of index instances. *)
+
+  val empty_cache : unit -> cache
+  (** Construct a new empty cache of index instances. *)
+
   val v :
     ?auto_flush_callback:(unit -> unit) ->
+    ?cache:cache ->
     ?fresh:bool ->
     ?readonly:bool ->
     log_size:int ->
@@ -127,6 +134,7 @@ module type S = sig
   (** The constructor for indexes.
 
       @param auto_flush_callback adds a callback before an auto flush.
+      @param cache a cache instance to use for instance sharing.
       @param fresh whether an existing index should be overwritten.
       @param read_only whether read-only mode is enabled for this index.
       @param log_size the maximum number of bindings in the `log` IO. *)
@@ -180,7 +188,7 @@ module type S = sig
 end
 
 module type Index = sig
-  (** The input of [Make] for keys. *)
+  (** The input of {!Make} for keys. *)
   module type Key = sig
     (* N.B. We use [sig ... end] redirections to avoid linking to the [_intf]
        file in the generated docs. Once Odoc 2 is released, this can be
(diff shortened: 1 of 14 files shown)


test/unix/common.ml
```diff
@@ -82,21 +82,24 @@ struct
   type t = {
     rw : Index.t;
     tbl : (string, string) Hashtbl.t;
+    cache : Index.cache;
```
icristescu, line 85: It does not matter too much for these tests (as each test use a fresh instance of the index), but before all tests used to shared the same cache. You could make the cache a global variable to simulate that behaviour. And there isn't any tests that uses the cache anyway, so no need to have it in the return type.

test/cache.ml
```diff
@@ -0,0 +1,31 @@
...
+
+let test_unbounded () =
+  let open Index.Cache.Unbounded in
+  (* Test that added entries are always found. *)
+  let c = create () in
+  find c "not-added" |> check_none "Cannot find non-existent value";
+  add c "added" 1;
+  find c "added" |> check_some "Can find added value" 1;
+  remove c "added";
+  find c "added" |> check_none "Can't find added value after remove";
```
icristescu, line 24:
```suggestion
  find c "added" |> check_none "Cannot find added value after remove";
```

test/cache.ml
```diff
@@ -0,0 +1,31 @@
...
+
+let test_noop () =
+  let open Index.Cache.Noop in
+  (* Test that added entries are never found. *)
+  let c = create () in
+  find c "not-added" |> check_none "Cannot find non-existent value";
+  add c "added" 1;
+  find c "added" |> check_none "Cannot find added value";
+  remove c "added";
+  find c "added" |> check_none "Can't find added value after remove";
```
icristescu, line 13:
```suggestion
  find c "added" |> check_none "Cannot find added value after remove";
```

src/index_intf.ml
```diff
@@ -114,9 +114,17 @@ module type S = sig
...
+  (** The type for caches of index instances. *)
+
+  val empty_cache : unit -> cache
+  (** Construct a new empty cache of index instances. *)
+
+  val v :
+    ?cache:cache -> ?fresh:bool -> ?readonly:bool -> log_size:int -> string -> t
   (** The constructor for indexes.
 
+      @param cache a cache instance to use for instance sharing.
```
icristescu, line 127:
```suggestion
      @param cache used for instance sharing.
```

test/unix/common.mli
```diff
@@ -22,9 +22,10 @@ module Index : Index.Private.S with type key = Key.t and type value = Value.t
 module Make_context (Config : sig
   val root : string
 end) : sig
-  type t = {
+  type t = private {
```
icristescu, line 25: why are you using private here?

test/unix/common.ml
CraigFe: > but before all tests used to shared the same cache. You could make the cache a global variable to simulate that behaviour.

I'd rather encourage the tests to be completely isolated and, as you say, it doesn't change existing semantics.

> And there isn't any tests that uses the cache anyway, so no need to have it in the return type.

Good point; removed. Not sure why I added it in the first place...

test/unix/common.mli
CraigFe: It's a type of values that is only _consumed_ in external code, never constructed – by design rather than coincidence, since there are internal invariants / log messages unavailable to consumers. I put `private` to make that "destruct only" intended usage explicit.

icristescu: I see, thanks!

samoht:
Could you rebase this PR?

CraigFe:
> Could you rebase this PR?

Done.

PR #189, opened 2020-06-29
auto flush callback

icristescu:
This PR adds a hook at the end of the merge, to give a change to `irmin-pack` to flush files to disk before the index does. This prevents invalid reads for RO in `irmin-pack`, where a key is found in index but not in pack.

On top of https://github.com/mirage/index/pull/175.

### DIFF
--- a/src/unix/index_unix.ml
+++ b/src/unix/index_unix.ml
@@ -39,10 +39,12 @@ module IO : Index.IO = struct
     mutable fan_size : int64;
     readonly : bool;
     buf : Buffer.t;
+    auto_flush_callback : unit -> unit;
   }
 
   let sync ?(with_fsync = false) t =
     if t.readonly then raise RO_not_allowed;
+    t.auto_flush_callback ();
     let buf = Buffer.contents t.buf in
     let offset = t.offset in
     Buffer.clear t.buf;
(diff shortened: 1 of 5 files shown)


icristescu:
As suggested by @CraigFe, I changed the base to master.

src/unix/index_unix.ml
```diff
@@ -151,9 +153,12 @@ module IO : Index.IO = struct
 
   let () = assert (String.length current_version = 8)
 
-  let v ~readonly ~fresh ~generation ~fan_size file =
+  let v ?auto_flush_callback ~readonly ~fresh ~generation ~fan_size file =
     let v ~fan_size ~offset raw =
       let header = 8L ++ 8L ++ 8L ++ 8L ++ fan_size in
+      let auto_flush_callback =
+        match auto_flush_callback with None -> Fun.id | Some f -> f
+      in
```
CraigFe, line 161:
```suggestion
  let v ?(auto_flush_callback = fun () -> ()) ~readonly ~fresh ~generation ~fan_size file =
    let v ~fan_size ~offset raw =
      let header = 8L ++ 8L ++ 8L ++ 8L ++ fan_size in
```

Not actually sure if it's better, but this is what I was imagining. As you prefer.

PR #191, opened 2020-07-06
Add minimal bench target

gs0510:
Running all the benchmarks is time intensive and also space-intensive.
These minimal benchmarks will be faster to run on PRs and since currently the
way to filter benchmarks is regex-based this commit introduces a new list to
run these minimal benchmarks.

### DIFF
--- a/bench/bench.ml
+++ b/bench/bench.ml
@@ -223,6 +223,7 @@ module Index = struct
     fresh : bool;
     benchmark : with_metrics:bool -> Index.t -> unit -> unit;
     dependency : string option;
+    speed : [ `Quick | `Slow ];
   }
 
   let suite =
@@ -234,6 +235,7 @@ module Index = struct
         fresh = true;
         benchmark = write_random;
         dependency = None;
+        speed = `Quick;
       };
       {
         name = "replace_random_with_timer";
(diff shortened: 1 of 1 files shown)


bench/bench.ml
```diff
@@ -316,6 +316,42 @@ module Index = struct
...
+      };
+      {
+        name = "find_random_rw";
+        synopsis = "[RW] Find in random order";
+        readonly = false;
+        fresh = false;
+        benchmark = find_random;
+        dependency = Some "replace_random";
+      };
+    ]
```
CraigFe, line 354: Quite a lot of duplication of metadata here. How about adding a record field ``speed = [ `Quick | `Slow ]`` and then filtering to pick only the `` `Quick`` tests in the case where `--minimal` is passed?

gs0510: That is a really good suggestion! Thanks!

bench/bench.ml
```diff
@@ -421,8 +435,15 @@ let print_json fmt (config, results) =
   in
   pretty_print fmt obj
 
+let get_suite_list minimal_flag =
+  if minimal_flag then
+    List.filter
+      (fun bench -> if bench.Index.speed = `Quick then true else false)
```
CraigFe, line 441:
```suggestion
      (fun bench -> bench.Index.speed = `Quick)
```

PR #192, opened 2020-07-08
Add an is_merging function

icristescu:
`is_merging` returns true is the index is running an async merge. 

I used the `merge_lock` to detect that a merge is ongoing. Other operations take this lock (`clear` and `close`), but as they are not async, so it should be fine.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -653,6 +653,11 @@ struct
         Thread.return `Completed
     | Some witness -> merge ?hook ~witness t
 
+  let is_merging t =
+    let t = check_open t in
+    if t.config.readonly then raise RO_not_allowed;
+    Mutex.is_locked t.merge_lock
+
   let replace t key value =
     let t = check_open t in
     Stats.incr_nb_replace ();
--- a/src/index.ml
+++ b/src/index.ml
@@ -653,6 +653,9 @@ struct
         Thread.return `Completed
     | Some witness -> merge ?hook ~witness t
 
+  (** [t.merge_lock] is used to detect an ongoing merge. Other operations can
+      take this lock, but as they are not async, we consider this to be a good
+      enough approximations. *)
   let is_merging t =
     let t = check_open t in
     if t.config.readonly then raise RO_not_allowed;
--- a/src/index_intf.ml
+++ b/src/index_intf.ml
@@ -83,8 +83,7 @@ module type MUTEX = sig
       unlocks [t]. *)
 
   val is_locked : t -> bool
-  (** Return true is the mutex is locked, but does not block the mutex
-      otherwise.*)
+  (** [is_locked t] returns [true] if the mutex is locked, without locking [t]. *)
 end
 
 module type THREAD = sig
@@ -176,7 +175,7 @@ module type S = sig
       {!RW_not_allowed} if called by a read-write index. *)
 
   val is_merging : t -> bool
-  (** [is_merging t] returns true if [t] is running an async merge. Raises
+  (** [is_merging t] returns true if [t] is running a merge. Raises
       {!RO_not_allowed} if called by a read-only index. *)
 end
 
--- a/src/index_intf.ml
+++ b/src/index_intf.ml
@@ -81,6 +81,10 @@ module type MUTEX = sig
   val with_lock : t -> (unit -> 'a) -> 'a
   (** [with_lock t f] first obtains [t], then computes [f ()], and finally
       unlocks [t]. *)
+
+  val is_locked : t -> bool
+  (** Return true is the mutex is locked, but does not block the mutex
+      otherwise.*)
 end
 
 module type THREAD = sig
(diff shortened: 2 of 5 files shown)


CraigFe:
Note that read-only instances don't share the same `merge_lock` as read-write instances, so they will always get `is_merging t = false` (except perhaps when a sibling in another thread is in the process of closing the instance).

icristescu:
> Note that read-only instances don't share the same `merge_lock` as read-write instances, so they will always get `is_merging t = false` (except perhaps when a sibling in another thread is in the process of closing the instance).

Indeed, I tried to clarify this point in the doc of `is_merging`. I added an exception if called by the RO instance to make this more clear.

CraigFe:
We now have quite a number of functions that work for only one of the instance types. By my count:

- **either**. `find`, `mem`, `close`, `iter`
- **RW-only**. `clear`, `flush`, `filter`, `replace`, `is_merging` (private: `force_merge` / `await`)
- **RO-only**. `sync`

And even the functions that _are_ shared often either change behaviour according to the instance type or do things that in both cases that only make sense in one, such as taking certain locks.

This is out-of-scope for this PR, but if this trend of diverging interfaces continuees I think we should consider splitting these out into different functors – or at least different in-memory representations for each instance type sharing the same functor (perhaps using phantom capabilities).

test/unix/force_merge.ml
```diff
@@ -338,6 +338,28 @@ let merge_during_sync () =
   Index.close rw;
   Index.close ro
 
+let test_is_merging () =
+  let Context.{ rw; _ } = Context.empty_index () in
+  let f msg b () =
+    Log.debug (fun l -> l "test ");
```
CraigFe, line 344:
```suggestion
    Log.debug (fun l -> l "test %s: expecting %b" msg b);
```

test/unix/force_merge.ml
```diff
@@ -350,4 +372,5 @@ let tests =
     ("find in async with log", `Quick, find_in_async_same_generation);
     ("sync and find after log cleared", `Quick, sync_after_clear_log);
     ("merge during ro sync", `Quick, merge_during_sync);
+    ("test is merging", `Quick, test_is_merging);
```
CraigFe, line 375:
```suggestion
    ("is_merging", `Quick, test_is_merging);
```

src/index_intf.ml
```diff
@@ -81,6 +81,10 @@ module type MUTEX = sig
   val with_lock : t -> (unit -> 'a) -> 'a
   (** [with_lock t f] first obtains [t], then computes [f ()], and finally
       unlocks [t]. *)
+
+  val is_locked : t -> bool
+  (** Return true is the mutex is locked, but does not block the mutex
+      otherwise.*)
```
CraigFe, line 87:
```suggestion
  (** [is_locked t] returns [true] if the mutex is locked, without locking [t]. *)
```

src/index.ml
```diff
@@ -651,6 +651,11 @@ struct
         Thread.return `Completed
     | Some witness -> merge ?hook ~witness t
 
+  let is_merging t =
+    let t = check_open t in
+    if t.config.readonly then raise RO_not_allowed;
+    Mutex.is_locked t.merge_lock
```
CraigFe, line 657: Perhaps add a comment here to say that we know that `merge_lock` isn't held only by the asynchronous merge process, but we're considering it a good enough approximation for this case.

test/unix/force_merge.ml
```diff
@@ -338,6 +338,28 @@ let merge_during_sync () =
...
+  let Context.{ rw; _ } = Context.empty_index () in
+  let f msg b () =
+    Log.debug (fun l -> l "test ");
+    Alcotest.(check bool) msg (Index.is_merging rw) b
+  in
+  f "before merge" false ();
+  let k1, v1 = (Key.v (), Value.v ()) in
+  Index.replace rw k1 v1;
+  let t = Index.force_merge ~hook:(before (f "before" true)) rw in
+  Index.await t |> check_completed;
```
CraigFe, line 351: This could be a helper e.g. `add_binding_and_merge ~hook`, to be used 3 times.

src/index_intf.ml
```diff
@@ -170,6 +174,10 @@ module type S = sig
   val sync : t -> unit
   (** [sync t] syncs a read-only index with the files on disk. Raises
       {!RW_not_allowed} if called by a read-write index. *)
+
+  val is_merging : t -> bool
+  (** [is_merging t] returns true if [t] is running an async merge. Raises
```
CraigFe, line 179:
```suggestion
  (** [is_merging t] returns true if [t] is running a merge. Raises
```

I don't think it technically has to be an async merge: it could be blocking in another thread owned by the user.

CraigFe:
- [ ] `CHANGES.md` entry

test/unix/force_merge.ml
icristescu: It's nicer with your suggestion, thanks!

PR #193, opened 2020-07-09
Add timer for merge

icristescu:
As merge is an async operation, I think we have to store a list of the last `n` merges. I picked the last 10 merges, but we can instead have `n` as parameter, if we think its better. 

Every merge operation is timed by default.

### DIFF
--- a/bench/bench.ml
+++ b/bench/bench.ml
@@ -93,6 +93,8 @@ module Benchmark = struct
     write_amplification_calls : float;
     write_amplification_size : float;
     replace_times : float list;
+    merge_times : float list;
+    nb_merges : int;
   }
   [@@deriving yojson]
 
@@ -113,6 +115,8 @@ module Benchmark = struct
     let ops_per_sec = nb_entriesf /. time in
     let mbs_per_sec = entry_sizef *. nb_entriesf /. 1_048_576. /. time in
     let replace_times = stats.replace_times in
+    let merge_times = stats.merge_times in
+    let nb_merges = stats.nb_merge in
     {
       time;
       ops_per_sec;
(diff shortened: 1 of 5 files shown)


src/stats.mli
```diff
@@ -16,6 +17,7 @@ type t = {
     - [bytes_written] is the number of bytes written to disk;
     - [nb_writes] is the number of writes to disk;
     - [nb_merge] is the number of times a merge occurred;
+    - [merge_times] lists how much time the last 10 merge operations took.
```
CraigFe, line 20:
```suggestion
    - [merge_times] lists how much time the last 10 merge operations took (in microseconds);
```

src/stats.ml
```diff
@@ -81,3 +84,13 @@ let sync_with_timer f =
   f ();
   let span = Mtime_clock.count timer in
   stats.time_sync <- Mtime.Span.to_us span
+
+let drop_head l =
+  match l with l when List.length l < 10 -> l | _ :: tl -> tl | [] -> []
```
CraigFe, line 89:
```suggestion
  if List.length l >= 10 then List.tl l else l
```

bench/bench.ml
```diff
@@ -132,10 +141,13 @@ module Benchmark = struct
        Read amplification in syscalls: %f@\n\
        Read amplification in bytes: %f@\n\
        Write amplification in syscalls: %f@\n\
-       Write amplification in bytes: %f" result.time result.ops_per_sec
-      result.mbs_per_sec result.read_amplification_calls
-      result.read_amplification_size result.write_amplification_calls
-      result.write_amplification_size
+       Write amplification in bytes: %f@\n\
+       Last 10 merges: %s@\n\
```
CraigFe, line 145:
```suggestion
       Last 10 merge durations (μs): %s@\n\
```

bench/bench.ml
```diff
@@ -93,6 +93,8 @@ module Benchmark = struct
     write_amplification_calls : float;
     write_amplification_size : float;
     replace_times : float list;
+    merge_times : float list;
```
CraigFe, line 96:
```suggestion
    merge_durations : float list; [@key "merge_durations_us"]
```

icristescu:
Thanks for your suggestions! I changed `times` with `durations` everywhere, it is better.

PR #195, opened 2020-07-11
New Index.clear semantics

samoht:
- `clear` now removes the index file(s)
- concurrent read-only instances see the changes only after a `sync`.
- When a read-only instance sync on a cleared file, it detects the change
  of generation and close/re-opam the file(s) on disk.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -100,7 +100,7 @@ struct
   let check_open t =
     match !t with Some instance -> instance | None -> raise Closed
 
-  let clear t =
+  let clear' t =
     let t = check_open t in
     Log.debug (fun l -> l "clear %S" t.root);
     if t.config.readonly then raise RO_not_allowed;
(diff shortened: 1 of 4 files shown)


CraigFe:
We seem to already have tests for "RO operations performed after clear", which haven't changed here. Are these new semantics observable to the user in any way?

samoht:
There is no observable difference from a user perspective in the current tests. Will try to come-up with new ones. Normally a read-only instance will always see the old results, even if the file is overwritten concurrently.

samoht:
Actually this is not needed: https://github.com/mirage/index/pull/198 shows that the current implementation respect snapshot isolation for read-only instances. This is the case because: the RO instance already keeps an open fd on an old merge file (which is removed by the next merge, so is not modified/overwritten) and keeps all the old log entries in memory, so it doesn't matter if we remove the log file or not.

PR #196, opened 2020-07-12
always fsync after a clear

samoht:


CraigFe:
Needs rebased over https://github.com/mirage/index/pull/197.

We should have a regression test for this. I believe something like "RW write `k`" → "RW clear" → "RO sync" → "RO read `k`" would expose this?

samoht:
I've added a regression tests but it showed that it was actually not very useful to call `flush` :-) we just want for `fsync` here, and I don't think we could test this properly.

PR #198, opened 2020-07-12
Add a test for snapshot isolation of read-only instances

samoht:


test/unix/main.ml
```diff
@@ -273,6 +273,28 @@ module Readonly = struct
     Index.close rw;
     Index.close ro
 
+  (* check that the ro instance is "snapshot isolated", e.g. it can read old
+     values, even when rw overwrites them. *)
+  let readonly_snapshot () =
+    let Context.{ rw; clone; tbl } = Context.full_index () in
+    let ro = clone ~readonly:true () in
+    check_equivalence rw tbl;
+    check_equivalence ro tbl;
```
CraigFe, line 282: `clone` and `full_index` should guarantee this (and if we want to test that, we can do it separately).

```suggestion
```

(We could also remove this from the test above.)

test/unix/main.ml
```diff
@@ -273,6 +273,28 @@ module Readonly = struct
...
+    let tbl2 =
+      let h = Hashtbl.create 103 in
```
CraigFe, line 284:
```suggestion
      let h = Hashtbl.create 0 in
```

I'm not sure why we picked 103 as the default size...

test/unix/main.ml
samoht: I think it's good to be explicit with the pre-conditions in this test

test/unix/main.ml
samoht: 103 is a good prime number :p

samoht:
Hum the test is failing on Travis :-/

CraigFe:
Interesting. Perhaps the flush needs a `with_fsync`?

samoht:
The missing key is the last one written:
```
- +1594567754us index [INFO] [index_19] merge
- +1594567754us index [DEBUG] [index_19] flushing instance
- +1594567754us index [INFO] [index_19] replace =a*DZ'*_~QDee~W;X^'d p_+AfKmq[.c<FdsPE?!E
- +1594567754us index [INFO] [index_19] replace ".kt4{.(y5m6H[f$5m_I |]?%Bv@6(+7c(?=@gNRO
- +1594567754us index [INFO] [index_19] replace UhvP]zSh\?Kq/3dzV*]P c?nj;7-pF7*mZ0)"UZn(
- +1594567754us index [INFO] [index_19] replace 7@\jm!6|!Nc,L5ay%06b j)|2E,6cK7&2mpR[i#dl
- +1594567754us index [INFO] [index_19] replace =P2d]'xpHO7.I~)K;[rB *(m?/!C@*]}~`([Yg}N4
- +1594567754us index_unix [DEBUG] [_tests/unix.main/index_19/index/merge] flushing 3800 bytes
- +1594567754us index_unix [DEBUG] [_tests/unix.main/index_19/index/log] set_header { offset = 0; generation = 40 }
- +1594567754us index_unix [DEBUG] [_tests/unix.main/index_19/index/log] flushing 200 bytes
- +1594567754us index_unix [DEBUG] [_tests/unix.main/index_19/index/log_async] set_header { offset = 0; generation = 40 }
- +1594567754us index [INFO] [index_19] merge
- +1594567754us index [DEBUG] [index_19] flushing instance
- +1594567754us index [INFO] [index_19] replace _;R/4Lh!)i#u([6)rf0e t]%%C._?f=W9W#TM].e4
- +1594567754us index [INFO] [index_19] replace 8HSVs#*Is|:r"C8:o1e` Yyw"KA][H:>6W8_U|Hk#
- +1594567754us index [INFO] [index_19] replace ON4]ZrETMUa;~z*eADeg f$+~uj0a=LkFi<7L>_o5
- +1594567754us index [INFO] [index_19] flush
- +1594567754us index [DEBUG] [index_19] flushing instance
- +1594567754us index_unix [DEBUG] [_tests/unix.main/index_19/index/log_async] flushing 120 bytes
```
It seems to be added to `log_async`, while there is a merge in progress. I suspect that the read-only instance doesn't know about that log_async file and so can't find the test. I need to check what happens when there is a race between `sync` and `merge`...

src/index.ml
```diff
@@ -388,10 +387,10 @@ struct
   let empty_cache = Cache.create
 
   let v ?(auto_flush_callback = fun () -> ()) ?(cache = empty_cache ())
-      ?(fresh = false) ?(readonly = false) ~log_size root =
+      ?(fresh = false) ?(readonly = false) ?(strict = false) ~log_size root =
```
CraigFe, line 390: We should make sure that this parameter is respected even when the RW instance is taken from the cache (which looks to be broken currently).

samoht:
Rebased the tests. Now that #200 has been fixed, this should be ok.

test/unix/main.ml
```diff
@@ -257,6 +257,56 @@ module Readonly = struct
...
+      m
+    in
+    let merge = lock () and sync = lock () in
+    let k, v = hashtbl_pick tbl2 in
+
+    Index.clear rw;
+    Index.replace rw k v;
+    let thread = Index.force_merge ~hook:(before merge sync) rw in
+    Hashtbl.iter (Index.replace rw) tbl2;
+    Index.flush ~with_fsync:true rw;
```
icristescu, line 302: We don't need `fsync` here.

PR #199, opened 2020-07-13
Scope test resource lifetimes inside continuations

CraigFe:
Change `Context.<>` functions to take a continuation representing the test.  After the test is complete, the main index instance and any clones created during the test will be closed. This avoids needing to explicitly `close` various indices for most tests.

This catches quite a few cases where test indices weren't being closed properly.

### DIFF
--- a/test/unix/main.ml
+++ b/test/unix/main.ml
@@ -1,6 +1,8 @@
 module I = Index
 open Common
 
+type index = Common.Index.t
+
 let ( // ) = Filename.concat
 
 let root = "_tests" // "unix.main"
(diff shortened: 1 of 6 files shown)


test/unix/main.ml
```diff
@@ -224,33 +204,35 @@ end
 (* Tests of read-only indices *)
 module Readonly = struct
   let readonly () =
-    let Context.{ rw; clone; _ } = Context.empty_index () in
+    let* Context.{ rw; clone; tbl; _ } = Context.with_empty_index () in
     let ro = clone ~readonly:true () in
-    Hashtbl.iter (fun k v -> Index.replace rw k v) main.tbl;
```
CraigFe, line 229: Unrelated change to get rid of an unnecessary `main` index instance here. Fortunately we're past those days :slightly_smiling_face:

Issue #200, opened 2020-07-14
race between sync and merge


samoht:
There is a race between `sync` and `merge` where some updates to `log_async` could not be seen by the read-only instance.

CraigFe:
Fixed in https://github.com/mirage/index/pull/203.

PR #203, opened 2020-07-14
Fix the race between merge and sync

samoht:
Related to #200

The failure is not fully reproducible, I still have no clue of what is happening.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,7 @@
 ## Fixed
 
 - Added values after a clear are found by read-only instances. (#168)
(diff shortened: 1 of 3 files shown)


samoht:
And here is a fix for #200

samoht:
Fix updated to not use an extra level of caching. We just need to compare the `log_async`'s generation with the "main" one to check whether is is outdated.

samoht:
Thanks @CraigFe for the very careful review. I'm adding a changelog entry and will merge once the CI is happy.

PR #204, opened 2020-07-14
Rename `IO.Header.{get_header,set_header}` into `{IO.Header.get,set}`

samoht:


samoht:
I've modified the changelog entry to also mention #175

PR #207, opened 2020-07-14
IO refactoring

samoht:
- Remove IO.set_generation, IO.get_generation and IO.force_offset
- Change IO.offset to take a `force` arguement
  (`IO.offset ~force:true` was `IO.force_offset`,
   `IO.offset ~force:false` was `IO.offset`)
- Add `IO.generation`. `IO.generation ~force:true` was `IO.get_generation`
  and add `IO.generation ~force:false` to return the cached generation number

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,10 +16,17 @@
   than waiting for it to finish. (#185)
 - `sync` has to be called by the read-only instance to synchronise with the
   files on disk. (#175)
-
 - Caching of `Index` instances is now explicit: `Index.Make` requires a cache
   implementation, and `Index.v` may be passed a cache to be used for instance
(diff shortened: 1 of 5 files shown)


CraigFe:
It's not clear to me why it's necessary to have a second layer of caching for the generation number, although perhaps I should first review https://github.com/mirage/index/pull/203 to see why it's useful there. (Will do that tomorrow.)

Looking at this PR purely by itself, I think it's a shame that IO takes on an extra responsibility in terms of managing shared state.

samoht:
Happy to have something different :-) But we need a way to easy way to track the IO metadata (offset, generation) of a file without to constantly reading the FS, and we need to check when they are out-dated.

PR #209, opened 2020-07-14
Add an option to choose overcommiting memory instead of blocking writes

### LINKED ISSUE #201: Allow to throttle writes when a merge is already in progress and that the caches are full
Currently if a merge is in progress and the caches are full, any call to replace` will be blocking. We should have an option (tentative name: `?throttle:['Block_writes | 'Overcommit_memory]`) to allow users to change this.

samoht:
This only happens when there is an async merge in progress and that the
caches are already full.

Fix #201 

This should probably be the default (and a `close` test should be changed to use `Block_writes`)

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -73,7 +73,14 @@ struct
 
   module Tbl = Hashtbl.Make (K)
 
-  type config = { log_size : int; readonly : bool; fresh : bool }
+  type throttle = [ `Overcommit_memory | `Block_writes ]
+
+  type config = {
+    log_size : int;
+    readonly : bool;
+    fresh : bool;
+    throttle : throttle;
+  }
 
   type index = { io : IO.t; fan_out : Fan.t }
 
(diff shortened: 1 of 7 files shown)


samoht:
Not that we can add a more clever option later on (e.g. `Delay_writes` where writes are delayed with an exponential back-off, depending on the memory already overcommitted)

src/index.ml
```diff
@@ -686,7 +698,11 @@ struct
           > 0)
     in
     if do_merge then
```
CraigFe, line 700: This variable probably wants renamed to something like `log_limit_reached`, since it's not determining whether we merge any more.

src/index_intf.ml
```diff
@@ -128,6 +128,7 @@ module type S = sig
     ?cache:cache ->
     ?fresh:bool ->
     ?readonly:bool ->
+    ?throttle:[ `Overcommit_memory | `Block_writes ] ->
```
CraigFe, line 131: Let's add an explanatory line in the doc-string (even though I think the type is very self-explanatory). At least, it makes sense to say what the default is.

CraigFe:
We also need a changelog entry for this one.

samoht:
Pushed your changes and a changelog. Any opinion on changing the default throttle behaviour?

CraigFe:
I think `` `Block_writes`` is the correct default to keep a bounded memory footprint, particularly since the distinction is rare in practice. (In the case where operations outpace the disk, I'd far rather block on finishing disk operations than keep a growing cache of TODOs that's going to need to be written eventually anyway.)

PR #211, opened 2020-07-15
Index.clear should cancel any concurrent Index.merge

samoht:
We want `clear` to be fast and it doesn't matter if we wait for any
async merges to complete as we will empty the entries anyway.

Fix #202

PR #212, opened 2020-07-15
Fix typo in clear

samoht:
We were clearing log twice, not log and log_async.


samoht:
As discussed offline, this is ready to merge. The issue found in #211 is related to `clear` cancelling `merges` and hitting a bug with shared write buffers.

PR #213, opened 2020-07-15
Index.rename should also rename the cached filename

samoht:
That doesn't change anything, but that makes the logs a bit prettier

### DIFF
--- a/src/unix/index_unix.ml
+++ b/src/unix/index_unix.ml
@@ -31,7 +31,7 @@ module IO : Index.IO = struct
   let ( -- ) = Int64.sub
 
   type t = {
-    file : string;
+    mutable file : string;
     mutable header : int64;
     mutable raw : Raw.t;
     mutable offset : int64;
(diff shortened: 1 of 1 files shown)


CraigFe:
What's the intended semantics of `rename`?

I would expect `~src` to be unusable after the rename has completed. (Or, for `~src` to be changed in place and to take a `~dst:string` rather than a `~dst:IO`.)

samoht:
Rename currently just changes `src` in place.

samoht:
I'm in favour of merging this one as it improves the debug messages (and will be needed if we ever want to unlink files). A proper API refactoring should probably happen later on.

CraigFe:
As mentioned offline, I think any use of `src` after the rename is likely to cause trouble. Where does this improve the logs? The only use I see is [here](https://github.com/mirage/index/blob/master/src/index.ml#L624), and the `src` isn't used after that point.

samoht:
It helped me to find the issue with the shares write buffers :-)

CraigFe:
CI failure unrelated. (The Ubuntu 20.04 runner doesn't support depexts properly.)

PR #215, opened 2020-07-16
Test abort merges after one entry added

icristescu:
This test reproduces the failing test of https://github.com/mirage/index/pull/214, but uses close instead of clear to abort the merge.

The fix is to [not cache buffers](https://github.com/mirage/index/pull/211/files#diff-24b617c6b01eb5122d1e4e45357860c6R180) which is already included in https://github.com/mirage/index/pull/211.

### DIFF
--- a/test/unix/force_merge.ml
+++ b/test/unix/force_merge.ml
@@ -366,6 +366,7 @@ let test_throttle () =
   Mutex.unlock m;
   Index.await thread |> check_completed
 
+(** Test that a clear is aborting the merge.*)
 let test_non_blocking_clear () =
   let* Context.{ rw; _ } = Context.with_empty_index () in
   let lock () =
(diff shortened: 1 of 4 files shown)


icristescu:
> Could you deduplicate the code in the two functions, e.g. by taking the aborting function as an argument?

Thanks, much cleaner with deduplication. Also clarified the comments.

test/unix/force_merge.ml
```diff
@@ -394,6 +395,44 @@ let test_non_blocking_clear () =
   | Ok `Aborted -> ()
   | _ -> Alcotest.fail "merge should have aborted"
 
+(** The test consists of aborting a first merge after one entry is added in the
+    ./merge file and checking that a second merge succeds. Regression test for
```
pascutto, line 399:
```suggestion
    ./merge file and checking that a second merge succeeds. Regression test for
```

PR #216, opened 2020-07-22
Improve behaviour and testing of `auto_flush_callback`

CraigFe:
Fixes various race conditions in the handling of `auto_flush_callback` and add regression tests.

In particular,
 - flushes of `log_async` now trigger the flush callback.
 - the callback is now triggered on explicit calls to `flush`, but not during `close` (previously, this was being handled inconsistently).
 - the callback _isn't_ triggered by flushes to the main Index IO, since we require that all values are first persisted in one of the two logs. (This was nearly the case before, and is now the case after fixing a race condition.)
 - added more documentation of the guarantees provided by the callback.


CraigFe:
Now also requiring OCaml 4.08, for a more convenient stdlib.

src/index.ml
```diff
@@ -633,8 +633,15 @@ struct
                   Tbl.replace log.mem key value;
                   append_key_value log.io key value)
                 log_async.mem;
-              IO.flush log.io;
+
+              (* IO.flush can fail due to an exception in [auto_flush_callback]. *)
+              (try IO.flush log.io
+               with exn ->
+                 Mutex.unlock t.merge_lock;
+                 raise exn);
```
pascutto: That should be handled by the `Mutex.with_lock` you're wrapped in, right?

CraigFe: True for the `rename_lock`, but not for the `merge_lock`. The `merge_lock` isn't properly scoped since it's taken by the main thread and dropped by the asynchronous one.

pascutto: Ah yes right, was too quick.

CraigFe:
Rebased.

CraigFe:
Pushed again to kick the CI over.

src/index_intf.ml
```diff
@@ -134,7 +134,14 @@ module type S = sig
     t
   (** The constructor for indexes.
 
-      @param auto_flush_callback adds a callback before an auto flush.
+      @param auto_flush_callback A function to be called before any new bindings
+      are persisted to disk (including both automatic flushing and explicit
+      calls to {!flush}). Values flushed during {!close} are excluded.
```
icristescu, line 139: What do you think of not calling the callback on explicit calls to flush? In irmin, we always call the callback explicitly when calling index.flush https://github.com/mirage/irmin/blob/master/src/irmin-pack/pack.ml#L195

CraigFe: There are two arguments _against_:

- Not calling the flush callback for explicit flushing requires the user to remember to do so (and presumably they will always have to do so if this is being used to maintain an invariant).
- It's convenient for testing if we have a way to trigger a flush in `log_async`, and without this we can only expose it by writing `1_000_000` bytes during a merge.

Arguments _for_ include:

- In cases of implementing an explicit flushing mechanism on top of this one (as in pack), it's perhaps more intuitive if the parent flushing is _not_ done via a callback.
-  `**auto**_flush_callback` is named to suggest that explicit flushes don't trigger it.

I see two solutions: (a) add some sort of other hook to fiddle with the `auto_flush_limit` so that we can test this properly, or (b) expose a switch on `Index.flush` (perhaps only in the private API for testing) that gives the user control over this behaviour.

What do you think?

pascutto: Indeed, that's what the name of the argument suggests, that requires to remove the `no_callback` argument in `flush` too.

Either that, or just name it `flush_callback`, which I tend to like more because it avoids the callback (as in https://github.com/mirage/irmin/blob/master/src/irmin-pack/pack.ml#L195). I can't see a reason why someone should have a different handling of automatic and manual calls to `flush`.

icristescu: I agree that `auto` strongly suggests that the callback is not called for explicit `index.flush` calls, and it makes sense to exclude both calls to `index.flush` and `index.close`.
And I like solution (b), seems simpler, but you can choose which one you prefer :)

PR #218, opened 2020-08-11
Ensure that non-blocking merges are triggered when overcommitting memory

CraigFe:
Fixes a bug where the log would grow without bound when using `` ~throttle:`Overcommit_memory ``, since implicit merges were avoided even when they would be non-blocking.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -727,10 +727,12 @@ struct
   (** [t.merge_lock] is used to detect an ongoing merge. Other operations can
       take this lock, but as they are not async, we consider this to be a good
       enough approximations. *)
+  let instance_is_merging t = Mutex.is_locked t.merge_lock
+
   let is_merging t =
     let t = check_open t in
     if t.config.readonly then raise RO_not_allowed;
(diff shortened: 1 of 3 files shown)


src/index.ml
```diff
@@ -727,10 +727,12 @@ struct
   (** [t.merge_lock] is used to detect an ongoing merge. Other operations can
       take this lock, but as they are not async, we consider this to be a good
       enough approximations. *)
+  let instance_is_merging t = Mutex.is_locked t.merge_lock
+
   let is_merging t =
     let t = check_open t in
     if t.config.readonly then raise RO_not_allowed;
```
pascutto, line 734: Totally unrelated, but the reason why we can't query a RO instance here is only because there is no convenient way to check that from a RO perspective, right?

CraigFe: As far as I understand it, yes.

Also, the motivation for this endpoint was to allow users to forego a `replace` if they knew that it would block. (So it was only deemed necessary for RW instances.)

pascutto: Got it, thanks!

PR #219, opened 2020-09-07
Simplify IO signature and provide an in-memory implementation

CraigFe:
This is a work-in progress towards providing an in-memory implementation of Index. This would serve as a reference for the semantics and help shape the IO signature into something that can reasonably be considered a "backend".

To begin with, this PR breaks up the current `IO` signature into two "platform" signatures (`IO` and `Lock`), and pulls out a separate `Thread` module as well. The `index` package will provide simple, in-memory instantiations of these signatures.

Tests are currently non-existent / broken.


CraigFe:
I've let this stagnate, so will close it for now. We can perhaps come back to this idea when making the shift to multicore.

PR #222, opened 2020-09-09
Use full unsigned int64 hashes

pascutto:
This PR removes the `hash_size` value in the input `Key` functor, and requires a `hash: t -> int64` function that is well distributed over all the 64 bits.
This `int64` value will always be interpreted as *unsigned*.

This breaks the on-disk format because it used these hashes to order the bindings in `data`. 
This means the version of the format should change and a migration function (that performs an external sort) should be provided.


pascutto:
The way to go would be to change the ordering instead, and use an ordering on the keys directly, rather than hashing them.

PR #224, opened 2020-09-10
Unvendor fsync and use Unix.fsync instead

pascutto:
`fsync` is available in the standard library since 4.08.


CraigFe (approved):

samoht:
Maybe `pread` and `pwrite` are the next good candidate for inclusion in the stdlib :p

PR #227, opened 2020-09-17
Test generation updates of log_async

icristescu:
PR https://github.com/mirage/index/pull/226 introduced a bug, reproduced by the test `sync_before_and_after_clearing_async`. The issue is when the end of a merge interleaves with two RO sync calls. 

It did not occur before #226 because the generation of log_async was [log's generation+1](https://github.com/mirage/index/pull/226/files#diff-732d381615307f94d510e2dbdee74f4aL683). This ensured that between the two RO sync there was a generation change for log_async, but lead to inconsistent generations between log and log_async. As mentioned in #226 this is a problem: the log_async file is read from scratch even when it is not necessary. 

One simple solution is to remove the [assert](https://github.com/mirage/index/pull/226/files#diff-732d381615307f94d510e2dbdee74f4aL683) and treat `offset > h.offset` as a generation change. I don't like this too much, because asserts help us discover these difficult bugs. Another possibility is to not use the generation of log async, but only the one of log. This is a more difficult change however, I'm not sure it works.

### DIFF
--- a/test/unix/force_merge.ml
+++ b/test/unix/force_merge.ml
@@ -297,6 +297,37 @@ let find_in_async_same_generation () =
   let t1 = Index.try_merge_aux ~force:true ~hook:(before f) w in
   Index.await t1 |> check_completed
 
+let sync_before_and_after_clearing_async () =
+  let* { Context.rw; clone; _ } = Context.with_full_index () in
+  let w = rw in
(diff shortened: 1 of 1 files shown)


icristescu:
Rebased. This PR adds a test for the `log_async` so we can merge it.

PR #229, opened 2020-09-24
Thread yields

icristescu:


pascutto:
Closing this as it doesn't seem to improve the repartition of resources, and introduces an overall performance regression.

PR #232, opened 2020-10-01
Move log_async clear under rename_lock.

pascutto:
This will prevent values from being added to the log before the log_async is cleared, as those values could possibly be discarded on next startup if a crash occured and the log_async was considered not empty.


CraigFe (approved):
Perhaps warrants a changelog entry as a minor bugfix.

PR #233, opened 2020-10-08
Add IO.exists

CraigFe:
... and remove calls to `Sys.file_exists` inside the `index` package, which is supposed to be platform-agnostic.

Also extracts namespacing choices to a new module, `Layout`. This will be more obviously useful when we add integrity-checking utilities that must know about the FS structure without necessarily relying on the entire implementation of Index.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -160,20 +160,6 @@ struct
     Mutex.with_lock t.rename_lock (fun () ->
         flush_instance ?no_callback ~with_fsync t)
 
-  let ( // ) = Filename.concat
-
-  let index_dir root = root // "index"
-
-  let log_path root = index_dir root // "log"
-
-  let log_async_path root = index_dir root // "log_async"
-
-  let index_path root = index_dir root // "data"
-
-  let lock_path root = index_dir root // "lock"
-
-  let merge_path root = index_dir root // "merge"
-
   let page_size = Int64.mul entry_sizeL 1_000L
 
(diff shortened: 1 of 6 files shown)


pascutto (approved):

src/index.ml
```diff
@@ -362,7 +348,7 @@ struct
     let generation =
       match log with None -> 0L | Some log -> IO.get_generation log.io
     in
-    let log_async_path = log_async_path root in
+    let log_async_path = Layout.log_async ~root in
     (* If we are in readonly mode, the log_async will be read during sync_log so
        there is no need to do it here. *)
     if (not readonly) && Sys.file_exists log_async_path then (
```
icristescu, line 354: you missed this one

icristescu: ah, you only want to use IO.exists in the index creation?

CraigFe: Indeed, good catch. Should be fixed.

pascutto: No it should be everywhere

icristescu (approved):

CraigFe:
There are two 60 minute timeout failures in the OCaml-CI jobs, but these are almost certainly unrelated as this is [already happening on `master`](https://ci.ocamllabs.io/github/mirage/index/commit/24075b167e9ec55c7ecf549842dd272cf7288e6d/variant/alpine-3.12-4.11).

Merging now.

PR #234, opened 2020-10-09
Make Index.close merge termination logic opt-in

CraigFe:
When developing the offline integrity-checking tool for Index, I noticed that the following simple usage of Index is non-deterministic:

```ocaml
let random () =
  let index = Index.v ~fresh:true ~log_size:100 "data/random" in
  for _ = 1 to 1001 do
    Index.replace index (random ()) (random ())
  done;
  Index.close index
```

In particular, the `Index.close` call may or may not cancel an ongoing merge operation, meaning that we cannot know what files will exist on disk after running this code. After pondering on it for a bit, I think this behaviour of `close` should be opt-in. This PR does exactly that.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,10 +19,12 @@
 - Add `IO.exists` obligation for IO implementations, to be used for lazy
   creation of IO instances. (#233, @CraigFe)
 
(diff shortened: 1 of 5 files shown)


icristescu (approved):
This looks good, I would just use `abort` instead of `immediately`, but both ways is fine.

PR #236, opened 2020-10-13
Add an initial integrity checking tool

CraigFe:
This adds a fairly basic integrity checking tool that allows any given `Index` module to become a `Cmdliner` binary serving offline statistics about stores built with those parameters. It has some rough edges and missing features (e.g. scan linearly for a specific key), but I think we can add those in later iterations.

There's quite a bit of code motion in this PR due to needing to extract various low-level logic from `index.ml` to places where it can more easily be used. Might be worth reviewing commit-by-commit.


src/checks.ml
```diff
@@ -0,0 +1,203 @@
...
+    include Io.Extend (IO)
+
+    (** This module never makes persistent changes *)
+    let v =
+      v ~fresh:false ~readonly:true ?flush_callback:None ~generation:0L
+        ~fan_size:0L
+
+    let iter ?min ?max f =
+      iter ?min ?max
+        ~page_size:Int64.(mul (of_int Entry.encoded_size) 1000L)
```
icristescu, line 16: To make this similar to the [`iter` in `index.ml`](https://github.com/mirage/index/pull/236/files#diff-488de9f5720c03d7021bc9e6e186bd0f516269ec43fe0f4cea0e82e7bd495063R79) I would define a separate `page_size` and have `f off entry`.

src/index.ml
```diff
@@ -18,6 +18,19 @@ all copies or substantial portions of the Software. *)
 include Index_intf
 module Stats = Stats
 module Cache = Cache
+module Checks = Checks
+
+module Key = struct
+  module type S = Key
+
+  module String_fixed = Data.String_fixed
```
icristescu, line 26: I think `String_fixed` is only used in tests, maybe we should define it only in tests?

src/index.ml
```diff
@@ -241,14 +216,14 @@ struct
         IO.v ~fresh:false ~readonly:true ~generation:0L ~fan_size:0L path
       in
       let mem = Tbl.create 0 in
-      iter_io (fun e -> Tbl.replace mem e.key e.value) io;
+      IO.iter (fun _ e -> Tbl.replace mem e.key e.value) io;
```
icristescu, line 219: Maybe you can add an `iter_io` function to make the diff a bit smaller.

src/index.ml
```diff
@@ -648,10 +628,11 @@ struct
               Tbl.clear log.mem;
               hook `After_clear;
               let log_async = assert_and_get t.log_async in
+              let append_io = IO.append log.io in
               Tbl.iter
                 (fun key value ->
                   Tbl.replace log.mem key value;
-                  append_key_value log.io key value)
+                  Entry.encode' key value append_io)
```
icristescu, line 635: Can you always inline the `append_io`? (In the call to `encode`, the `IO.append` sometimes it's inlined and sometimes it's not.)

test/cli/dune
```diff
@@ -0,0 +1,32 @@
...
+
+(executable
+ (name generate)
+ (modules generate)
+ (libraries index index.unix))
+
+(rule
+ (alias generate-cli-test-data)
+ (targets data)
+ (package index)
```
icristescu, line 14: I'm not sure I understand why we need this option; why would this alias be build outside of the package?

src/checks.ml
```diff
@@ -0,0 +1,203 @@
...
+            ( quick_stats $ setup_log,
+              Term.info ~doc:"Print high-level statistics about the store."
+                "stat" );
+            ( integrity_check $ setup_log,
+              Term.info
+                ~doc:"Search the store for integrity faults and corruption."
+                "check" );
+          ]
+        |> (exit : unit result -> _));
+      assert false
```
icristescu, line 199: Why do you need this assert false? Also, can you use `unit` instead of `empty` here ?

test/cli/stat.t/run.t
```diff
@@ -0,0 +1,37 @@
...
+  
+  {
+    "entry_size": "40.0 B",
+    "files": {}
+  }
+
+Running `stat` on an index after 10 merges:
+
+  $ ../index_fsck.exe stat ../data/random > log.txt 2>&1
+  $ sed -re 's/"lock": "[0-9]+"/"lock": "<PID>"/g' log.txt
```
icristescu, line 14: `sed --r` does not work on macOS (`sed: illegal option -- r`), but `sed -E ..` does.

icristescu (approved):
Thanks, this looks great, and I learned a few things while reviewing it.

src/checks.ml
CraigFe: The `assert false` and `empty` are both used to make it clear at the type-level that this function never returns. Returning `unit` would also work, but doesn't make that explicit.

test/cli/dune
CraigFe: It's to distinguish it from the `index-bench` package, which doesn't need to generate this data as part of its build process.

src/index.ml
CraigFe: We'd need to define it in two places in the tests (`test/unix` and `test/cli`), which is what pushed me into putting it here instead. I also think it makes sense to provide a default implementation here, but am happy to go back to the previous state if you disagree.

src/checks.ml
CraigFe: The `assert false` and `empty` are both used to make it clear at the type-level that this function never returns. Returning `unit` would also work, but doesn't make that explicit.

test/cli/dune
CraigFe: It's to distinguish it from the `index-bench` package, which doesn't need to generate this data as part of its build process.

src/index.ml
CraigFe: Not inlining here saves allocating a closure for each entry in the table (and similarly for the `iter_io`), but I can still inline it if you prefer?

src/index.ml
icristescu: Having a default implementation make sense, it's fine with me to keep them here.

src/index.ml
CraigFe: Yep, thanks.

src/checks.ml
icristescu: I see, thanks!

src/checks.ml
CraigFe: Yep, thanks.

src/index.ml
icristescu: Ah, I missed that. Keep it as is it then.

test/cli/dune
icristescu: Thanks!

PR #237, opened 2020-10-13
Add capabilities to the fanout

pascutto:
The `Fan` module only works with the following workflow:
- Do a lot of `update`
- Do a `finalize`
- Do a lot of `search`

This PR makes this workflow appear in the types and ensures the invariant is verified.﻿

### DIFF
--- a/src/fan.mli
+++ b/src/fan.mli
@@ -15,37 +15,37 @@ furnished to do so, subject to the following conditions:
 The above copyright notice and this permission notice shall be included in
 all copies or substantial portions of the Software. *)
 
-type t
+type 'a t
 
-val equal : t -> t -> bool
-(** The equality function for fan-out. *)
+val equal : 'a t -> 'a t -> bool
+(** The equality function for fanouts. *)
 
-val v : hash_size:int -> entry_size:int -> int -> t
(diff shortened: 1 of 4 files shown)


src/fan.mli
```diff
@@ -15,37 +15,37 @@ furnished to do so, subject to the following conditions:
...
-val update : t -> int -> int64 -> unit
+val update : [ `Write ] t -> int -> int64 -> unit
 (** [update t hash off] updates [t] so that [hash] is registered to be at offset
     [off]. *)
 
-val finalize : t -> unit
-(** Finalizes the update of the fanout. This is mendatory before any [search]
+val finalize : [ `Write ] t -> [ `Read ] t
+(** Finalizes the update of the fanout. This is mandatory before any [search]
     query. *)
```
CraigFe, line 39:
```suggestion
(** Finalizes the update of the fanout. *)
```

Not really needed now that the types imply it, but keep it if you want.

CraigFe (approved):
LGTM.

src/fan.mli
pascutto: Indeed, but I thought it wasn't that costly to keep it there. Happy to remove it if that feels odd though.

CraigFe: Np; let's keep it.

PR #239, opened 2020-10-21
Fix race condition in Close.abort_merge test

CraigFe:
The asynchronous merging thread must wait for the parent to signal the merge before it can proceed, otherwise it might finish the merge before the parent thread gets a chance to issue the abort.


icristescu (approved):
Seems a good fix to me.

PR #241, opened 2020-10-27
Use Semaphore compatibility library

CraigFe:
Our usage of the `Mutex` module in the UNIX implementation of Index previously relied upon undefined behaviours of `glibc`, which caused Index to be incompatible with FreeBSD. In particular, we were relying on being able to release a mutex from a thread that did not itself take the mutex (in order to implement the merge lock).

This commit replaces the usage of `Mutex` with a dependency on the `semaphore-compat` library, which provides a shim of the `Semaphore` module to be released in OCaml 4.12. In particular, the mutual-exclusion semantics that we require are provided by the binary semaphores in this library.

### DIFF
--- a/src/index_intf.ml
+++ b/src/index_intf.ml
@@ -34,7 +34,9 @@ module type SEMAPHORE = sig
   (** The type of binary semaphore. *)
 
   val make : bool -> t
-  (** Return a fresh semaphore with the given initial state. *)
+  (** [make b] returns a new semaphore with the given initial state. If [b] is
+      [true], the semaphore is initially available for acquisition; otherwise,
+      the semaphore is initially unavailable. *)
 
   val acquire : t -> unit
   (** Acquire the given semaphore. Acquisition is not re-entrant. *)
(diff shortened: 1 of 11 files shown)


icristescu (approved):
Thanks, looks good!

src/index_intf.ml
```diff
@@ -27,28 +27,29 @@ end
...
+  (** Binary semaphores for mutual exclusion *)
 
   type t
-  (** The type of mutual-exclusion locks. *)
+  (** The type of binary semaphore. *)
 
-  val create : unit -> t
-  (** Return a fresh mutex. *)
+  val make : bool -> t
+  (** Return a fresh semaphore with the given initial state. *)
```
icristescu, line 37: can you specify what the initial state means, i.e. `true` for released and `false` for acquired?

CraigFe: Good idea; done.

CraigFe:
I can confirm that this fixes compatibility with FreeBSD :tada:

PR #244, opened 2020-11-13
io: use a list of buffers instead of a Buffer.t to cache IO writes

samoht:
Trying to test the new CB :-)


CraigFe:
Looks like the CB has made it's verdict :-)

image

Shall we close this PR, or do you want to modify the approach @samoht?

samoht:
Let's close this one; I'd be interested to know if bindings to `pwritev` would make this go faster.

PR #246, opened 2020-11-17
index: optimize code layout if the hot path

samoht:


samoht:
CB seems to agree with me as the results are better than on master ; can someone approve and merge that PR?

CraigFe (approved):

src/index.ml
```diff
@@ -441,28 +441,31 @@ struct
...
-          t.log_async
-        @~ find_log_index)
+    let find_index () =
+      find_if_exists ~name:"index" ~find:interpolation_search t.index
+    in
+    Semaphore.with_acquire t.rename_lock @@ fun () ->
+    match find_log_async () with
+    | e -> e
+    | exception Not_found -> (
+        match find_log () with e -> e | exception Not_found -> find_index ())
```
samoht, line 468: I think I've introduced a bug here as there is one less `Not_found` catch handler.

PR #247, opened 2020-11-17
index: allocate less

samoht:


samoht:
Added one more commit.

samoht:
CB seems happy here as well!

CraigFe (approved):

PR #251, opened 2020-12-01
change info message into debug to tune down verbose outputs

samoht:

PR #253, opened 2020-12-07
Add an overcommit option in replace to postpone merges

icristescu:
Experimenting with adding more control over merges from the freeze thread in irmin-pack. For now the irmin-pack benchmarks look promising:
- the spikes are the freezes that block waiting for previous freeze to complete
freezes
- if we remove the spike due to freezes and only look at the time for the rest of the commits:
wo_freezes

On top of https://github.com/mirage/index/pull/251.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -549,7 +549,7 @@ struct
       incr n;
       !n
 
-  let merge ?(blocking = false) ?(filter = fun _ -> true) ?(hook = fun _ -> ())
+  let merge' ?(blocking = false) ?(filter = fun _ -> true) ?(hook = fun _ -> ())
       ~witness t =
     let yield () = check_pending_cancel t in
     Semaphore.acquire t.merge_lock;
@@ -715,7 +715,9 @@ struct
     | None ->
         Log.debug (fun l -> l "[%s] index is empty" (Filename.basename t.root));
         Thread.return `Completed
-    | Some witness -> merge ?hook ~witness t
+    | Some witness -> merge' ?hook ~witness t
+
+  let merge t = ignore (force_merge ?hook:None t : _ async)
 
   (** [t.merge_lock] is used to detect an ongoing merge. Other operations can
       take this lock, but as they are not async, we consider this to be a good
@@ -754,7 +756,7 @@ struct
           None
       | `Overcommit_memory, false | `Block_writes, _ ->
           let hook = hook |> Option.map (fun f stage -> f (`Merge stage)) in
-          Some (merge ?hook ~witness:(Entry.v key value) t)
+          Some (merge' ?hook ~witness:(Entry.v key value) t)
     else None
 
   let replace ?overcommit t key value =
@@ -778,7 +780,7 @@ struct
     | None ->
         Log.debug (fun l -> l "[%s] index is empty" (Filename.basename t.root))
     | Some witness -> (
-        match Thread.await (merge ~blocking:true ~filter:f ~witness t) with
+        match Thread.await (merge' ~blocking:true ~filter:f ~witness t) with
         | Ok (`Aborted | `Completed) -> ()
         | Error (`Async_exn exn) ->
             Fmt.failwith "filter: asynchronous exception during merge (%s)"
--- a/src/index.ml
+++ b/src/index.ml
@@ -727,7 +727,7 @@ struct
     if t.config.readonly then raise RO_not_allowed;
     instance_is_merging t
 
-  let replace' ?hook t key value =
+  let replace' ?hook ?(overcommit = false) t key value =
     let t = check_open t in
     Stats.incr_nb_replace ();
     Log.debug (fun l ->
@@ -745,7 +745,8 @@ struct
           Tbl.replace log.mem key value;
           Int64.compare (IO.offset log.io) (Int64.of_int t.config.log_size) > 0)
     in
-    if log_limit_reached then
+    if overcommit then None
+    else if log_limit_reached then
       let is_merging = instance_is_merging t in
       match (t.config.throttle, is_merging) with
       | `Overcommit_memory, true ->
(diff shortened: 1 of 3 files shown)


samoht:
Can you also plot the memory usage to see the impact on the memory usage? (I suspect this won't be much but it's worth checking)

icristescu:
Here is the maxrss for these benchmarks
maxrss_irmin_pack

src/index_intf.ml
```diff
@@ -169,6 +169,9 @@ module type S = sig
   (** [is_merging t] returns true if [t] is running a merge. Raises
       {!RO_not_allowed} if called by a read-only index. *)
 
+  val force_merge : t -> unit
```
CraigFe, line 172: - Can we just call it `merge`? I think it's only called `force_merge` internally because we already had a function called `merge`.
- I think I would probably expect this to block; perhaps worth mentioning that it doesn't in the doc-comment.

src/index_intf.ml
```diff
@@ -128,7 +128,7 @@ module type S = sig
   val mem : t -> key -> bool
   (** [mem t k] is [true] iff [k] is bound in [t]. *)
 
-  val replace : t -> key -> value -> unit
+  val replace : ?overcommit:bool -> t -> key -> value -> unit
   (** [replace t k v] binds [k] to [v] in [t], replacing any existing binding of
       [k]. *)
```
CraigFe, line 133: `?overcommit` probably wants a mention in the doc-comment. I'd also prefer `?overcommit:unit`, but perhaps I'm the only one.

src/index.ml
```diff
@@ -745,7 +747,8 @@ struct
           Tbl.replace log.mem key value;
           Int64.compare (IO.offset log.io) (Int64.of_int t.config.log_size) > 0)
     in
-    if log_limit_reached then
+    if overcommit then None
+    else if log_limit_reached then
```
CraigFe, line 751:
```suggestion
    if log_limit_reached && not overcommit then
```

src/index_intf.ml
icristescu: Good point, can you check if the doc is clearer now? https://github.com/mirage/index/pull/253/files#diff-7f70b3697164504072732d126e841e5a06b5456874bebb65fead19a71cd639baR176

CraigFe: Looks great, thanks :+1:

CraigFe (approved):

src/index_intf.ml
```diff
@@ -169,6 +172,10 @@ module type S = sig
   (** [is_merging t] returns true if [t] is running a merge. Raises
       {!RO_not_allowed} if called by a read-only index. *)
 
+  val merge : t -> unit
+  (** [merge t] forces a merge for [t]. It is non-blocking, i.e. it returns
+      immediately, with the merge running concurrently. *)
```
pascutto, line 177: It does block if a merge is already running, right?

icristescu: Indeed. I changed it, let me know if it's clear

icristescu:
I added a commit with the changes proposed by `opam-dune-lint` as the linting job was failing. However, if I add the   `"ppx_repr" {>= dev}` to `index-bench.opam` as suggested by `opam-dune-lint`, there is a dependency cycle 
```
ppxlib -> ppxlib.0.17.0
    ppx_repr 0.1.0 requires >= 0.12.0 & < 0.18.0
```
I'm not sure how to fix

samoht:
@icristescu it should be fixed by #1199

src/index_intf.ml
icristescu: I changed it back to bool option, to make it more consistent with irmin-pack arguments (especially here https://github.com/mirage/irmin/pull/1183/files#diff-7e9a5902736413eca81eab1399b2e2af88d7b0ed48dd90597a9873f0bcc42730R61 I think it does not look nice to mix unit and bool optional arguments).

CraigFe: It makes sense to me to have a mix of the two when certain flags have a clear default and others do not, but as you prefer.

talex5:
@icristescu : `opam-dune-lint` is new and might be buggy. What was the full error you got? The bit you pasted doesn't look like an error: it's just saying it chose `ppxlib.0.17.0`, and warning that isn't the latest version but it had to pick it due to a requirement from `ppx_repr`. The actual error will be elsewhere.

icristescu:
I added back the problematic commit, the logs now should show the problem.

> The bit you pasted doesn't look like an error: it's just saying it chose `ppxlib.0.17.0`, and warning that isn't the latest version but it had to pick it due to a requirement from `ppx_repr`. The actual error will be elsewhere.

Sorry about that, that was the last bit before the error, so I wrongly suspected it to be the problem.

talex5:
OK, so the error is:

```
- index-bench -> (problem)
    Rejected candidates:
      index-bench.dev: Requires ppx_repr >= dev
```

Perhaps you just need to remove the `>= dev` constraint? It defaults to using the version you are using as the lower bound, but maybe you don't really need `dev`? The lint error in CI suggests:

```
index-bench.opam: changes needed:
  "ppx_repr" {>= 0.1.0}
```

icristescu:
@talex5 thanks a lot, that fixed the issue

icristescu:
Rebased. If no objection, I'll merge this shortly.

src/index_intf.ml
```diff
@@ -128,9 +128,12 @@ module type S = sig
...
   (** [mem t k] is [true] iff [k] is bound in [t]. *)
 
-  val replace : t -> key -> value -> unit
+  val replace : ?overcommit:bool -> t -> key -> value -> unit
   (** [replace t k v] binds [k] to [v] in [t], replacing any existing binding of
-      [k]. *)
+      [k].
+
+      If [overcommit] is passed, the operation does not triger a merge, even if
+      the caches are full. *)
```
samoht, line 136: Maybe somehow add that  the default is `overcommit=false`?

PR #254, opened 2020-12-07
Add log messages at the end of merges

CraigFe:
As requested by @icristescu in https://github.com/mirage/index/pull/251.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -543,11 +543,19 @@ struct
     in
     (go [@tailcall]) true 0L 0 0
 
+  let merge_counter =
+    let n = ref 0 in
+    fun () ->
+      incr n;
+      !n
+
   let merge ?(blocking = false) ?(filter = fun _ -> true) ?(hook = fun _ -> ())
       ~witness t =
     let yield () = check_pending_cancel t in
     Semaphore.acquire t.merge_lock;
-    Log.info (fun l -> l "[%s] merge" (Filename.basename t.root));
+    let merge_id = merge_counter () in
+    Log.info (fun l ->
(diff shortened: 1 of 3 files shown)


samoht:
Can you also tune down the `Log.info` for `flush` and `sync` which are very verbose?

samoht:
Might also be useful to add the duration of a merge when it ends. (and make sure there is no extra remaining `Log.info` for merges)

CraigFe:
Done and done; thanks.

samoht:
Can you display the duration in seconds? :-)

CraigFe:
(Done :slightly_smiling_face:)

src/index.ml
```diff
@@ -543,11 +543,20 @@ struct
...
+      incr n;
+      !n
+
   let merge ?(blocking = false) ?(filter = fun _ -> true) ?(hook = fun _ -> ())
       ~witness t =
     let yield () = check_pending_cancel t in
     Semaphore.acquire t.merge_lock;
-    Log.info (fun l -> l "[%s] merge" (Filename.basename t.root));
+    let merge_id = merge_counter () in
+    let merge_start_time = Mtime_clock.counter () in
```
icristescu, line 557: there is also this timer for merge https://github.com/mirage/index/blob/master/src/index.ml#L668; maybe it's not worth keeping both? (you can remove the one in stats)

CraigFe: I avoided doing that originally since I wanted to keep the statistics about merge times, but it's fairly easy to just pass the merge duration to `Stats` instead. I've pushed a commit that does that :+1:

icristescu: Thanks! merge whenever you're ready

icristescu (approved):

PR #260, opened 2020-12-18
index: Index.iter should only iter once over the entries

samoht:


samoht:
I've stolen the tests from #181 and the seem to pass too.

icristescu:
The tests pass because we are using `Tbl.replace` which ensures that we don't have duplicates. I added these tests in https://github.com/mirage/index/pull/181 to check that we don't break the invariant with `Tbl.add`. But regardless of what we do with https://github.com/mirage/index/pull/181, I think we can merge these tests.

samoht:
Yes what I meant is that even if we don't iterated over the index twice the tests still pass (and thus iterating over the index twice sounds suspicious).

icristescu (approved):

PR #264, opened 2020-12-21
Avoid resizing hashtables too often

samoht:
The log hashbtbl is expected to hold [config.log_size] elements,
so initialise these with the right size at the first place.


samoht:
I'm undecided about this one.

samoht:
Closing as it slows down the irmin testrun quite a bit: from a few ms to run a single test to more than 10s 😱

PR #266, opened 2020-12-21
Fix tests

samoht:
Not sure why this changed...


samoht:
This CI doesn't like the change (as expected) so should be an issue with my local checkout..

Issue #268, opened 2021-01-18
Test failure: assertion failed on version header read


CraigFe:
We got the following (somewhat ominous) failure in the Irmin CI:

```
        test alias test/irmin-pack/runtest (exit 1)
(cd _build/default/test/irmin-pack && ./test.exe -q --color=always)
Testing `irmin-pack'.
This run has ID `B4A75E2B-8612-4B45-A0B0-DDB36A2D7957'.

  [OK]          PACK                     0   Basic operations on contents.
  [OK]          PACK                     1   Basic operations on nodes.
  [OK]          PACK                     2   Basic operations on commits.
  [OK]          PACK                     3   Basic operations on branches.
  [OK]          PACK                     4   Hash operations on trees.
  [OK]          PACK                     5   Basic merge operations.
  [OK]          PACK                     6   Basic operations on slices.
  [OK]          PACK                     7   Test merges on tree updates.
  [OK]          PACK                     8   Tree caches and hashconsing.
  [OK]          PACK                     9   Complex histories.
  [OK]          PACK                    10   Empty stores.
  [OK]          PACK                    11   Private node manipulation.
  [OK]          PACK                    12   High-level store operations.
  [OK]          PACK                    13   High-level operations on trees.
  [OK]          PACK                    14   High-level store synchronisation.
  [OK]          PACK                    15   High-level store merges.
  [OK]          PACK                    16   Unrelated merges.
  [OK]          PACK                    17   Low-level concurrency.
  [OK]          PACK                    18   Concurrent updates.
  [OK]          PACK                    19   with_tree strategies.
  [OK]          PACK                    20   Concurrent head updates.
  [OK]          PACK                    21   Concurrent merges.
  [OK]          PACK                    22   Shallow objects.
  [OK]          PACK                    23   Closure with disconnected commits.
  [OK]          PACK                    24   Clear.
  [OK]          PACK                    25   Graph.Iter.
  [OK]          PACK                    26   Watch.Callbacks and exceptions.
  [OK]          PACK                    27   Watch.Basic operations.
  [OK]          LAYERED_PACK             0   Basic operations on branches.
  [OK]          LAYERED_PACK             1   Basic merge operations.
  [OK]          LAYERED_PACK             2   Complex histories.
  [OK]          LAYERED_PACK             3   Empty stores.
  [OK]          LAYERED_PACK             4   Basic operations on slices.
  [OK]          LAYERED_PACK             5   Private node manipulation.
[...]

PR #269, opened 2021-01-21
Improve information given by failing assert statements in IO

CraigFe:
Should help in diagnosing https://github.com/mirage/index/issues/268.

### DIFF
--- a/src/unix/raw.ml
+++ b/src/unix/raw.ml
@@ -65,23 +65,34 @@ let unsafe_read t ~off ~len buf =
   Stats.add_read n;
   n
 
+let assert_read ~len n =
+  assert (
+    if Int.equal n len then true
+    else (
+      Printf.eprintf "Attempted to read %d bytes, but got %d bytes instead!\n%!"
+        len n;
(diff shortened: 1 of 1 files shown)


src/unix/raw.ml
```diff
@@ -65,33 +65,45 @@ let unsafe_read t ~off ~len buf =
   Stats.add_read n;
   n
 
+let assert_read ~len n =
+  assert (
+    if Int.equal n len then true
+    else (
+      Printf.eprintf "Attempted to read %d bytes, but got %d bytes instead!\n%!"
+        len n;
```
icristescu, line 73: Can you also print the backtrace?

CraigFe: I'd rather enable backtrace collection in all tests, if we can do that: https://github.com/mirage/irmin/pull/1254

This way we take care of potential future hard-to-reproduce exceptions too.

icristescu: Great 👍

icristescu (approved):

PR #272, opened 2021-02-16
Report fanout size in index-fsck

icristescu:


src/checks.ml
```diff
@@ -72,8 +77,12 @@ module Make (K : Data.Key) (V : Data.Value) (IO : Io.S) = struct
     let io path =
       with_io path @@ fun io ->
       let IO.Header.{ offset; generation } = IO.Header.get io in
+      let fanout_size =
+        let size = IO.get_fanout_size io |> Int64.to_int in
```
CraigFe, line 81: It's unlikely to be an issue, but for the sake of reporting the file contents faithfully it'd be best to avoid doing `Int64.to_int` here. Perhaps `Bytes` should be taking an `int64` now instead?

samoht:
@CraigFe : can you merge this one if you are happy with the changes?

CraigFe (approved):
Thanks @icristescu :-)

PR #273, opened 2021-02-16
index: use 63-bit integers in the fanout

CraigFe:
On 64-bit systems, this should reduce the memory usage of the fan-out by a factor of 4.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+# Unreleased
+
+## Changed
+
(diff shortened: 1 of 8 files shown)


icristescu (approved):
Thanks, this is great, merge when ready.
 
For future reference I'm reposting here the memtrace graphs before and after this PR.
memtrace_initial
memtrace_int63

CraigFe:
I've dropped the pin-depend, since Optint is now released with the changes we need. Will merge when the CI is happy, which may take a little while.

PR #276, opened 2021-02-19
Internal documentation

pascutto:
This adds some documentation throughout of the code. The codebase is quite complicated to get into, and not documented at all.
It's not finished yet, but I expect we can iterate a little bit before merging, so reviewing is already welcome.
I'm happy to remove some comments if they are too verbose or noisy.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -32,15 +32,14 @@ module Value = struct
   module String_fixed = Data.String_fixed
 end
 
-let may f = function None -> () | Some bf -> f bf
-
-let assert_and_get = function None -> assert false | Some e -> e
-
 exception RO_not_allowed
+(** Raised whenever a read-only instance performs a write action. *)
 
 exception RW_not_allowed
+(** Raised whenever a read-write instance performs a reserved read-only action. *)
 
 exception Closed
+(** Raised whenever a closed instance is used. *)
 
 module Make_private
     (K : Key)
@@ -79,9 +78,8 @@ struct
   module IO = struct
     include Io.Extend (IO)
 
-    let page_size = Int64.mul entry_sizeL 1_000L
-
     let iter ?min ?max f =
+      let page_size = Int64.mul entry_sizeL 1_000L in
       iter ~page_size ?min ?max (fun ~off ~buf ~buf_off ->
           let entry = Entry.decode buf buf_off in
           f off entry;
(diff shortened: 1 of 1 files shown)


src/index.ml
```diff
@@ -32,15 +32,14 @@ module Value = struct
   module String_fixed = Data.String_fixed
 end
 
-let may f = function None -> () | Some bf -> f bf
-
-let assert_and_get = function None -> assert false | Some e -> e
-
 exception RO_not_allowed
+(** Raised whenever a RO instance performs a RW-only action. *)
```
icristescu, line 36: we are using `read-write` and `read-only` in the index_intf docs, so maybe we should keep that convention.

src/index.ml
```diff
@@ -93,42 +91,73 @@ struct
...
+    log_size : int;  (** The log maximal size before triggering a [merge]. *)
     readonly : bool;
     fresh : bool;
-    throttle : throttle;
+        (** Whether the index was created from scratch, erasing existing data,
+            if any. *)
+    throttle : throttle;  (** The throttle strategy used when merges stack. *)
     flush_callback : unit -> unit;
+        (** A callback called whenever the index is flushed. Useful to sync
+            index's flushes with larger systems. *)
```
icristescu, line 102:
```suggestion
            index's flushes with the flushes of the larger systems. *)
```

src/index.ml
```diff
@@ -93,42 +91,73 @@ struct
...
   }
 
-  type index = { io : IO.t; fan_out : [ `Read ] Fan.t }
+  type index = {
+    io : IO.t;  (** The disk file handler. *)
+    fan_out : [ `Read ] Fan.t;
+        (** The fan-out, used to map keys to small intervals in the file, in
```
icristescu, line 108: I would add that it is both in-memory and on disk.

src/index.ml
```diff
@@ -93,42 +91,73 @@ struct
...
+    io : IO.t;  (** The disk file handler. *)
+    mem : value Tbl.t;  (** The in-memory counterpart of [io]. *)
+  }
 
   type instance = {
     config : config;
-    root : string;
+    root : string;  (** The root location of the index *)
     mutable generation : int64;
+        (** The index generations is a counter of rewriting operations (e.g.
```
icristescu, line 121:
```suggestion
        (** The generation is a counter of rewriting operations (e.g.
```

src/index.ml
```diff
@@ -93,42 +91,73 @@ struct
...
+            [clear] and [merge]). It is used to sync RO instances. *)
     mutable index : index option;
+        (** The main index file contains old sorted bindings. It is [None] when
+            no [merge] occurred yet. On RO instances, this is also [None] when
+            the file is empty, e.g. after a [clear]. *)
     mutable log : log option;
+        (** The log file contains the most recent bindings. It can be [None]
+            when a RO in created before its RW counterpart; in that case the
```
icristescu, line 129:
```suggestion
            when a RO is created before its RW counterpart; in that case the
```

src/index.ml
```diff
@@ -179,36 +210,11 @@ struct
...
-          (* Convert fractional position to position in output space *)
-          let position = low_out +. (proportion *. (high_out -. low_out)) in
-          let rounded = ceil (position -. 0.5) +. 0.5 in
-          Int64.of_float rounded
-      end)
+  (** {1 RO instances syncing} *)
 
+  (** Loads the log file at [path], if it exists. Used by RO instances, when
+      they have been created before their RW counterpart, to fill the [log]
+      field, or to load the temporary [log_async]. *)
```
icristescu, line 217: I think it loads the log_async regardless of when RO is created, but the phrase suggests otherwise. can you rephrase?

src/index.ml
```diff
@@ -222,41 +228,55 @@ struct
...
+  (** Syncs the [log_async] of the instance by checking on-disk changes. *)
   let sync_log_async t =
     match t.log_async with
     | None -> t.log_async <- try_load_log t (Layout.log_async ~root:t.root)
     | Some log ->
         let offset = IO.offset log.io in
         let h = IO.Header.get log.io in
+        (* If the generation has changed, reload everything. *)
         if t.generation <> h.generation then sync_log_entries log
+          (* else if the valid offset is greater, reload the newest data. *)
```
icristescu, line 246:
```suggestion
          (* else if the disk offset is greater, reload the newest data. *)
```
maybe use disk offset and instance/cached offset throughout?

src/index.ml
```diff
@@ -222,41 +228,55 @@ struct
...
         else if offset < h.offset then sync_log_entries ~min:offset log
+          (* else if the offset is lesser, that means the [log_async] was
+             cleared, and the generation should have changed. *)
         else if offset > h.offset then assert false
 
-  let sync_index ~generation t =
-    may (fun (i : index) -> IO.close i.io) t.index;
+  (** Syncs the [index] of the instance by checking on-disk changes.
+      [generation] *)
```
icristescu, line 253:
```suggestion
   *)
```

src/index.ml
```diff
@@ -222,41 +228,55 @@ struct
...
+  let sync_index t =
+    (* Close the file handler to be able to reload it, as the file Option.iter have
```
icristescu, line 255:
```suggestion
    (* Close the file handler to be able to reload it, as the file has
```

src/index.ml
```diff
@@ -748,10 +845,13 @@ struct
 
   let try_merge t = ignore (try_merge_aux ?hook:None ~force:false t : _ async)
 
-  (** [t.merge_lock] is used to detect an ongoing merge. Other operations can
-      take this lock, but as they are not async, we consider this to be a good
-      enough approximations. *)
-  let instance_is_merging t = Semaphore.is_held t.merge_lock
+  (** {1 Replace} *)
+
+  let instance_is_merging t =
```
icristescu, line 850: maybe it makes more sense to have this in the `Merges` section?

icristescu (commented):
Thanks, I have just a few comments.

src/index.ml
pascutto: You mean instead of `RO` and `RW`?

pascutto:
Fixed, thanks!

samoht (approved):

PR #277, opened 2021-02-20
merge: only decode the value when necessary

pascutto:


icristescu:
I run memtrace with this patch on top of https://github.com/mirage/index/pull/273 and `filter` is a the top allocator:
Screenshot 2021-02-22 at 13 45 49
(the line number is incorrect because its on my branch https://github.com/icristescu/index/tree/CraigFe-fanout-63-bit-integers/).

It seems that `~some` allocates some space even if its not used. Do you mind inlining the filter in the `if` condition?

icristescu (approved):
Thanks!

CraigFe:
The benchmarking UI seems to be suggesting that this introduces a 4% – 12% performance regression depending on the benchmark and metric, but I'm not sure if that's within variance or not.

pascutto:
> The benchmarking UI seems to be suggesting that this introduces a 4% – 12% performance regression depending on the benchmark and metric, but I'm not sure if that's within variance or not.

I don't know how to explain this.
- The 4% regressions happen in read-only benchmarks that do not perform any merges, so the code I modified is not accessed at all. This is likely to be explained by the bench variance.
- The 12% regressions on the `merge_durations_us` are unlikely to be due to the normal variance. I'm not sure how this PR can introduce such a regression though. I would have expected the opposite variation as this PR avoids (during a regular merge):
  - Filtering the `log` at the beginning of the `merge`.
  - Unnecessarily decoding the `values` of the `index` bindings

icristescu:
The merge durations reported in the benchmarks also include the time the merge thread yields to the main thread. If we want to accurately measure merges, we should maybe add a benchmark where the merges are blocking?

PR #278, opened 2021-02-25
Remove writing buffers

pascutto:
This seems to be a big perf regression with my homemade benchs; but checking with the upstream ones.

### DIFF
--- a/src/unix/index_unix.ml
+++ b/src/unix/index_unix.ml
@@ -26,64 +26,41 @@ let current_version = "00000001"
 module Stats = Index.Stats
 
 module IO : Index.IO = struct
-  external ( ++ ) : int64 -> int64 -> int64 = "%int64_add"
-
-  external ( -- ) : int64 -> int64 -> int64 = "%int64_sub"
+  let ( ++ ) = Int64.add
 
   type t = {
     mutable file : string;
     mutable header : int64;
     mutable raw : Raw.t;
     mutable offset : int64;
-    mutable flushed : int64;
     mutable fan_size : int64;
     readonly : bool;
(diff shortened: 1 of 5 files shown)


src/unix/index_unix.ml
```diff
@@ -26,64 +26,41 @@ let current_version = "00000001"
 module Stats = Index.Stats
 
 module IO : Index.IO = struct
-  external ( ++ ) : int64 -> int64 -> int64 = "%int64_add"
-
-  external ( -- ) : int64 -> int64 -> int64 = "%int64_sub"
+  let ( ++ ) = Int64.add
```
samoht, line 29: this is allocating too much (keep the externals)

src/unix/index_unix.ml
```diff
@@ -26,64 +26,41 @@ let current_version = "00000001"
...
-    Raw.close t.raw
-
-  let auto_flush_limit = 1_000_000L
+  let close t = Raw.close t.raw
 
   let append t buf =
+    t.flush_callback ();
     if t.readonly then raise RO_not_allowed;
-    Buffer.add_string t.buf buf;
+    Raw.unsafe_write t.raw ~off:(t.header ++ t.offset) buf;
```
samoht, line 58: This will cause a lot of short writes. Are you sure it's ok?

pascutto: It's not, I want to benchmark how much it is not :)

samoht: sounds good :-)

pascutto:
Experiment finished, sorry for the noise :)

PR #279, opened 2021-02-25
Do not allocate when flushing

pascutto:


samoht:
This needs to cleaning up - @pascutto will do it one day :-)

CraigFe:
Rebased.

src/unix/buffer.ml
```diff
@@ -0,0 +1,31 @@
+open! Import
```
samoht, line 1: missing headers

src/unix/raw.ml
```diff
@@ -34,9 +34,9 @@ let fsync t = Unix.fsync t.fd
 let close t = Unix.close t.fd
 let fstat t = Unix.fstat t.fd
 
-let unsafe_write t ~off buf =
+let unsafe_write t ~off buf buf_offset length =
```
icristescu, line 37: nitpicking, but do you mind renaming `buf_offset` to `buffer_offset` as above?

src/unix/raw.ml
```diff
@@ -85,7 +85,7 @@ module Version = struct
     assert_read ~len n;
     Bytes.unsafe_to_string buf
 
-  let set t v = unsafe_write t ~off v
+  let set t v = unsafe_write t ~off v 0 (String.length v)
```
icristescu, line 88: we can replace `String.length v` with `8` here I think

icristescu (approved):
Looks good, thanks! can you add some memtrace graphs to track the effect of this PR?

CraigFe:
Thanks! Merging.

This is a very significant performance improvement for write-intensive loads. Running benchmarks with:

```bash
; dune exec bench/bench.exe -- --nb-entries 10_000_000 --filter '^replace_random$' --log-size 200_000
```

For this benchmark, this PR reduces total allocations from 35G to 24G, and makes memory usage more stable. Before:

image

After:

image

Issue #285, opened 2021-03-22
Use `int63`s for file offsets


pascutto:
Index counterpart of https://github.com/mirage/irmin/issues/1330.

Using `int63`s instead of `int64`s has shown great result in https://github.com/mirage/index/pull/273.
We should investigate the impact of changing all `int64`s to `int63`s for file offsets in the `irmin-pack` stack.

PR #286, opened 2021-03-23
Use 63-bit integers as file offsets

CraigFe:
Fix https://github.com/mirage/index/issues/285.

Currently using an unreleased feature in `optint` to allow observing the type equality without a coercion. Very slightly faster than using the (identity) cast, which doesn't seem to be inlined w/o Flambda 2.

### DIFF
--- a/src/unix/index_unix.ml
+++ b/src/unix/index_unix.ml
@@ -26,17 +26,17 @@ let current_version = "00000001"
 module Stats = Index.Stats
 
 module IO : Index.IO = struct
-  external ( ++ ) : int64 -> int64 -> int64 = "%int64_add"
+  let ( ++ ) = Int63.add
 
-  external ( -- ) : int64 -> int64 -> int64 = "%int64_sub"
+  let ( -- ) = Int63.sub
 
   type t = {
     mutable file : string;
-    mutable header : int64;
+    mutable header : Int63.t;
     mutable raw : Raw.t;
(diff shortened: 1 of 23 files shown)


src/unix/pwrite.c
```diff
@@ -4,7 +4,26 @@
 #include <caml/signals.h>
 #include <caml/unixsupport.h>
 
-CAMLprim value caml_index_pwrite
+CAMLprim value caml_index_pwrite_int
```
CraigFe, line 7: I tried to use macros to avoid this duplication of the bindings, but it doesn't seem to be worth it.

src/unix/index_unix.ml
```diff
@@ -26,17 +26,17 @@ let current_version = "00000001"
 module Stats = Index.Stats
 
 module IO : Index.IO = struct
-  external ( ++ ) : int64 -> int64 -> int64 = "%int64_add"
+  let ( ++ ) = Int63.add
```
CraigFe, line 29: Could (and perhaps should) inline this again using `Int63.observe`.

CraigFe: On second thought, this isn't inline-able anyway since the 32-bit representation isn't semantically `int64`.

CraigFe:
Necessary `optint` feature now merged (thanks @dinosaure :pray:), so merging.

PR #288, opened 2021-03-26
Remove files on disk on `clear`

icristescu:
After a `clear` the files on disk should only contain the header  

For now this only works (and is tested) for the `data` file.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -990,6 +990,7 @@ module Private = struct
   module Io_array = Io_array
   module Search = Search
   module Data = Data
+  module Layout = Layout
 
   module Hook = struct
     type 'a t = 'a -> unit
--- a/src/index.ml
+++ b/src/index.ml
@@ -164,7 +164,8 @@ struct
         t.pending_cancel <- false;
         t.generation <- Int63.succ t.generation;
         let log = Option.get t.log in
-        IO.clear ~generation:t.generation log.io;
+        let hook () = hook `IO_clear in
+        IO.clear ~generation:t.generation ~hook log.io;
         Tbl.clear log.mem;
         Option.iter
           (fun l ->
--- a/src/index.ml
+++ b/src/index.ml
@@ -235,9 +235,13 @@ struct
     | Some log ->
         let offset = IO.offset log.io in
         let h = IO.Header.get log.io in
-        (* If the generation has changed, reload everything. *)
-        if t.generation <> h.generation then sync_log_entries log
-          (* else if the disk offset is greater, reload the newest data. *)
+        (* If the generation has changed *)
+        if t.generation <> h.generation then (
+          (* close the file .*)
+          IO.close log.io;
+          (* check that file is on disk, reopen and reload everything. *)
+          t.log_async <- try_load_log t (Layout.log_async ~root:t.root)
+          (* else if the disk offset is greater, reload the newest data. *))
         else if offset < h.offset then sync_log_entries ~min:offset log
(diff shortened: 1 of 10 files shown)


icristescu:
I separated `IO.v` in two: `IO.v` for read-write instances and `IO.v_readonly` for readonly ones, to ensure that the readonly instances can interact with `clear`:
- the readonly instance always tries to open the file - in this way we avoid the two steps of (1) checking if the file is on disk, and (2) opening the file and failing because the file was removed in the meantime;
- the readonly instance always tries to read the header - if the header is not written then the readonly should assume the file is not yet available. (this change should also help with the flaky test in https://github.com/mirage/index/issues/268).

But there is still a race between `sync` and `clear`exposed in the test [readonly_io_clear](https://github.com/mirage/index/pull/288/commits/34f00151a10090d7ae617a6f0e75673f64ff032f#diff-00a2414af7b93f0c86283b3e01a854bb69b7a598259b023470271033ff8a431bR301) in the last commit; `clear` (1) sets the generation, and (2) removes the file - if between (1) and (2) readonly does `sync`, then readonly has the new generation but the old file and misses the new updates even if it calls `syncs`afterwards

samoht:
We want to run 2h on an archive mode on this one to check for regressions.

icristescu:
Rebased. 
The new clear implementation is a bit tricky so I did it in several steps: 
- the more straightforward implementation (as done in irmin-pack) is in commit 44ed68e5c97204b7b3f5927c188578b178904339 and consists of (1) set the new generation in the file (2) close and remove file; (3) open a new file with the new generation. 
- this implementation leads to a race between readonly sync and clear exposed in the test in commit 2d6eb61ab6853d21f4b57d0c5fe3c7cd2f304234 (it can occur in irmin-pack too [here](https://github.com/mirage/irmin/blob/master/src/irmin-pack/IO.ml#L157)): if the sync occurs between steps (1) and (2) of clear, then this is the old file that is reopened by the readonly instance instead of the (yet non-existing) new file.
- the fix proposed by @gbelouze is commit 0f98e721650cd76301c43de4ae79b80cf5001cb7 and consists in exploiting the fact that the readonly’s fd to a file does not change even if we rename the file. Then clear: (1) renames file into a temporary file (2) opens new file with the new generation; (3) set new generation in temporary file; (4) close and remove temporary file. This trick ensures that the readonly only sees a generation change when the new file is already available on disk.

samoht:
LGTM could you rebase it?

samoht:
@icristescu can you run `dune build @fmt` ?

@CraigFe a quick review would be great :-)

CraigFe (approved):

src/index.ml
```diff
@@ -235,9 +236,13 @@ struct
...
         let h = IO.Header.get log.io in
-        (* If the generation has changed, reload everything. *)
-        if t.generation <> h.generation then sync_log_entries log
-          (* else if the disk offset is greater, reload the newest data. *)
+        (* If the generation has changed *)
+        if t.generation <> h.generation then (
+          (* close the file .*)
+          IO.close log.io;
+          (* check that file is on disk, reopen and reload everything. *)
+          t.log_async <- try_load_log t (Layout.log_async ~root:t.root)
```
samoht, line 244: For the record: this is likely to be slow when there is a blocked merge but this is similar to what was done before (e.g #296) ; we'll work on a fix in a separate PR.

samoht (approved):

PR #291, opened 2021-04-01
Specialise `v` to readonly and readwrite instances

icristescu:
extracted from https://github.com/mirage/index/pull/288.

I'm still not sure how the flaky tests in https://github.com/mirage/index/issues/268 and https://github.com/mirage/irmin/issues/1245 occur, but one possible explanation is that the readonly tries to read the header  of a file while that header is not yet written. This can happen if the file is being created simultaneously to a readonly trying to do a sync. 

This PR introduces a `v_readonly` that 
- eliminates the two steps (1) check if file exists and (2) then open it (the file might disappear between the two depending on how we implement `clear` );
- allows readonly to read the header - if the header is not yet written then readonly "acts" as if the file is not yet created.

I'm not sure we want to do this though (and we don't really need it for https://github.com/mirage/index/pull/288) but it might prevent some flaky tests from occurring.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Changed
 
+- Specialise `IO.v` to create read-only or read-write instances. (#291)
+
 - Optimised the in-memory representation of index handles, resulting in a
   significant reduction in memory use. (#273)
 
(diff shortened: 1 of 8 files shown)


CraigFe:
LGTM; nice to drop one of these flags.

I think it's a little odd to have `v_readonly` return `t option` and not either `result` or a `t` w/ exception, but don't have a strong opinion.

icristescu:
> I think it's a little odd to have `v_readonly` return `t option` and not either `result` or a `t` w/ exception, but don't have a strong opinion.

I opted for returning `result`, I agree its nicer than an `option` type.

samoht:
@CraigFe : can you merge if you are happy with the changes? This would unlock #288

CraigFe (approved):

PR #292, opened 2021-04-02
Updated the benches to allow multiple executions

mattiasdrp:
The displayed result is the mean of all the computed results

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,8 @@
 - Optimised the in-memory representation of index handles, resulting in a
   significant reduction in memory use. (#273)
 
+- Benches are now executed 3 times and a new option `nb-exec` has been added (#292)
+
 # 1.3.0 (2021-01-05)
 
 ## Added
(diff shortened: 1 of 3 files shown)


samoht:
It's a good idea to add more iterations to reduce variance but it would be nice to not have to wait for 30 minutes to get a feedback. We can remove one or two existing benchmarks to avoid this.

It would also be a good idea to use mtime instead of systime (in a different PR).

mattiasdrp:
Testing with 

```bash
dune exec -- ./bench/bench.exe --json --minimal --nb-exec n
```
| `n` | time |
| ------------- | ------------- |
| 1 |  155,22s = 3 min |
| 3 | 436,05s = 7 min |
| 5  | 763,81s = 12min  |

Do you have some preferences? The difference between 1 and 3 executions is not that huge

icristescu (approved):

icristescu:
can you rebase it?

mattiasdrp:
Done

icristescu:
Thanks!

PR #297, opened 2021-04-13
expose header size

gbelouze:
Expose in `IO` a function to get the size of the header of `data`. This necessary to implement a migration protocol from `data` towards another storing format (namely, for btrees).

### DIFF
--- a/src/io_intf.ml
+++ b/src/io_intf.ml
@@ -35,6 +35,9 @@ module type S = sig
 
   val close : t -> unit
 
+  val size_header : t -> int
+  (** [size_header t] is [t]'s header size. *)
+
   module Lock : sig
     type t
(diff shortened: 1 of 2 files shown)


src/io_intf.ml
```diff
@@ -35,6 +35,9 @@ module type S = sig
 
   val close : t -> unit
 
+  val size_header : t -> int
+  (** Returns the size of the header. *)
```
samoht, line 39:
```suggestion
  (** [size_header t] is [t]'s header size. *)
```

samoht: (a pure function doesn't return anything, it *is* a value)

icristescu:
Thanks!

PR #298, opened 2021-04-16
Reproduce flaky test

icristescu:
This test reproduces the flaky test in https://github.com/mirage/irmin/issues/1233.  It is due to a race between readonly sync and (the end) of merge:
- RW adds `v1` in log and calls merge;
- RW adds `v2` in log_async;
- RO calls sync - it reads the offset and generation of the log but does not update its in-memory log yet ;
- end of merge: the offset and generation of the log are updated, the contents of log_async are moved in the log;
- RO's sync continues - it reads the log expecting `v1` but it reads `v2` instead; `v1` is lost for RO;

I think this is what happens because in the logs we have this:
```
index           [DEBUG] [upper1] sync
+3995024us index           [DEBUG] [upper1] checking for changes on disk
+3995044us index           [DEBUG] [upper1] checking on-disk log_async file
+3995119us index_unix      [DEBUG] [_build/layered_4/upper1/index/log] get_headers: { offset = 165; generation = 6 }
----- here the sync is interrupted for the end of the merge ----
us index_unix      [DEBUG] [_build/layered_4/upper1/index/log] set_header +3995147+3995175us index          
m+3995273us irmin.pack.io   [DEBUG
m+3995324us irmin.pack.io   [DEBUG] _build/layered_4/upper1/store.pack
---- log async contents are moved to log ----
IO flush +3995366us index_unix      [] [_build/layered_4/upper1/index/log] flushing 33 bytes
m+3995420us index_unix      [DEBUG [_build/layered_4/upper1/index/log_async] set_header { offset = 0; generation = 8] [ }
index           [INFO [upper1] merge completed  
```
 
This is fixed by https://github.com/mirage/index/pull/288 because the clear operation called at the end of the merge creates a new file - so the old one can still be used by the RO to update its in memory log.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,5 +1,7 @@
-version = 0.17.0
+version = 0.18.0
 profile = conventional
+
 break-infix = fit-or-vertical
 parse-docstrings = true
 indicate-multiline-delimiters = no
+module-item-spacing = compact
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,11 +1,21 @@
 # Unreleased
 
+## Fixed
+
+- Fixed a crash-consistency bug due to a potential flush of an incomplete entry
+  to disk. Entries are now flushed as complete strings. (#301)
+
 ## Changed
 
 - Specialise `IO.v` to create read-only or read-write instances. (#291)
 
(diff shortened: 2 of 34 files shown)


samoht:
Merged as part of #288

PR #300, opened 2021-04-22
Replay bootstrapping trace

icristescu:
Initial benchmarks for replaying the trace of index operations during bootstrap. The trace is small for now, I’m working on getting a bigger one.

It needs the dependency on `tezos-context-hash` but only on the Hash module, and it needs to be in a separate package (https://github.com/icristescu/tezos-context-hash/tree/hash_pkg); Otherwise, we end up with the following dune error:

```
Error: Conflict between the following libraries:
- "index" in _build/default/src
- "index" in /Users/icristes/.opam/4.11.0/lib/index
  -> required by library "irmin-pack" in
     /Users/icristes/.opam/4.11.0/lib/irmin-pack
  -> required by library "tezos-context-hash" in
     _build/default/vendors/tezos-context-hash/src
-> required by executable replay in bench/dune:25

```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "vendors/tezos-context-hash"]
+	path = vendors/tezos-context-hash
+	url = https://github.com/tarides/tezos-context-hash
--- a/bench/replay.ml
+++ b/bench/replay.ml
@@ -1,0 +1,261 @@
+(*
+ * Copyright (c) 2018-2021 Tarides <contact@tarides.com>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *)
+
+open Common
+
+module Encoding = struct
+  module Hash = Tezos_context_hash.Hash
+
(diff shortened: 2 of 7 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "vendors/tezos-context-hash"]
+	path = vendors/tezos-context-hash
+	url = https://github.com/icristescu/tezos-context-hash
```
Ngoguey42, line 3: I suppose this should be set to the right URL before merging

bench/replay.ml
```diff
@@ -0,0 +1,266 @@
...
+                  in
+                  if b <> b' then
+                    Fmt.failwith "Operation find %a expected %b got %b"
+                      (Repr.pp Encoding.Key.t) k b b'
+              | Add (k, v) ->
+                  let k = key_to_hash k in
+                  Store.replace store k v
+              | Ro_mem _ | Ro_find _ -> ()
+            in
+            progress Int64.one;
```
Ngoguey42, line 199: This is going to be rendered a lot. It might currently slow down the replay.

Try adding a `~sampling_interval` to `Progress_unix.counter`. Aim for 1 to 100 renders per sec.

bench/replay.ml
```diff
@@ -0,0 +1,266 @@
...
+
+module Index = struct
+  module Index =
+    Index_unix.Make (Encoding.Key) (Encoding.Val) (Index.Cache.Unbounded)
+  include Index
+
+  let cache = Index.empty_cache ()
+
+  let v root =
+    Index.v ~cache ~readonly:false ~fresh:false ~log_size:500_000 root
```
Ngoguey42, line 149: Isn't the store `fresh`?

Ngoguey42 (commented):
👍

bench/replay.ml
CraigFe: (Aside: I will fix this tediousness in the next release of Progress...)

samoht:
This is blocked on @CraigFe's reorganisation of tezos-context-hash into different libraries.

CraigFe:
Thanks! I've pushed a suggested alternative to the Hash extraction [here](https://github.com/tarides/tezos-context-hash/pull/26), which is very similar to yours, and PRed the necessary changes to this patch [here](https://github.com/icristescu/index/pull/2).

CraigFe:
I'm not sure why the CI is failing to install `digestif` for `tezos-context-hash`; will ask around.

index-bench.opam
```diff
@@ -1,30 +1,32 @@
...
   "ppx_deriving_yojson"
   "re"
   "stdlib-shims"
   "yojson"
   "ppx_repr"
+  "logs" {>= "0.7.0" & with-test}
+  "optint" {>= "0.1.0" & with-test}
+  "progress" {>= "0.1.1" & with-test}
+  "repr" {>= "0.2.1" & with-test}
+  "rusage" {>= "1.0.0" & with-test}
```
kit-ty-kate, line 25:
```suggestion
  "rusage" {>= "1.0.0" & with-test}
  "digestif" {>= "0.9.0"}
```

CraigFe (approved):
Thanks all :-) The CI is now happy, so I propose we merge this.

PR #301, opened 2021-04-26
Crash-consistency bug fix

mattiasdrp:
```ocaml
(* Entry *)
let encode' key value f =
  f (K.encode key);
  (* Flush possible. Crash-consistency bug *)
  f (V.encode value)
```

Called like this:

```ocaml
Entry.encode' key value (IO.append t)
```
with
```ocaml
let append t buf =
  Buffer.add_string t.buf buf;
  flush t (* if offset is high enough *)
```

This means that the encoding function takes a writing function `f` and applies it to the encoding. If this function `f` flushes a flush may occur and if a crash happens at the same time the flushed encoding will be incomplete (the key will have been written but not the value)

The easiest way to prevent it is to call `f` on the concatenation of the two encoding since this won't induce any API change.

Since `Repr.encode` will change in a near future, the priority is to fix this bug and work on the encoding once `Repr.encode` has been changed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Unreleased
 
+## Fixed
+
+- Fixed a crash-consistency bug due to a potential flush of an incomplete entry
+  to disk. Entries are now flushed as complete strings. (#301)
+
 ## Changed
 
 - Specialise `IO.v` to create read-only or read-write instances. (#291)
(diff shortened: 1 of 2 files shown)


CraigFe (approved):
Looks good. It's a performance hit, but we can sort that out later as you say. Needs a CHANGES entry.

mattiasdrp:
Done, ready to be merged if it's ok for you

CraigFe:
Thanks; merging.

PR #303, opened 2021-04-27
[experimental] Add a heuristic to balance main and merging threads

pascutto:


samoht:
That seems useful:

Screenshot 2021-04-28 at 14 52 41

src/index.ml
```diff
@@ -656,6 +665,20 @@ struct
...
+       execution. Relies on `yield` being called for each merging page. *)
+    let nb_merge_yields = div index_size (of_int merge_page_size) in
+    if nb_merge_yields = zero then None
+    else
+      (* This is the number of replaced entries between each `yield` necessary
+         such that the merging thread and the main thread call `yield` the same
+         number of time. *)
+      let yield_frequency = div log_size nb_merge_yields in
+      (* We leave some margin for the merging thread to finish. *)
+      to_int (div yield_frequency (of_int 2)) |> Option.some
```
Ngoguey42, line 680: Let's make sure the result of that division is not zero

src/index.ml
```diff
@@ -876,8 +903,16 @@ struct
...
+  let yield t =
+    let log = Option.get t.log in
+    (* yield every `t.replace_yield_frequency`, when it exists. *)
+    Option.iter
+      (fun i -> if Tbl.length log.mem mod i = 0 then Thread.yield ())
+      t.replace_yield_frequency
+
   let replace' ?hook ?(overcommit = false) t key value =
     let t = check_open t in
+    yield t;
```
Ngoguey42, line 915: Since this function looks like `Lwt_unix.auto_yield`, we could name it the name way.

```suggestion
  let auto_yield t =
    let log = Option.get t.log in
    (* yield every `t.replace_yield_frequency`, when it exists. *)
    Option.iter
      (fun i -> if Tbl.length log.mem mod i = 0 then Thread.yield ())
      t.replace_yield_frequency

  let replace' ?hook ?(overcommit = false) t key value =
    let t = check_open t in
    auto_yield t;
```

src/index.ml
```diff
@@ -876,8 +903,16 @@ struct
 
   (** {1 Replace} *)
 
+  let yield t =
+    let log = Option.get t.log in
+    (* yield every `t.replace_yield_frequency`, when it exists. *)
+    Option.iter
+      (fun i -> if Tbl.length log.mem mod i = 0 then Thread.yield ())
```
Ngoguey42, line 910: What if `Tbl.length log.mem` didn't change between 2 calls to `yield`?

Ngoguey42 (commented):
I think the overall idea is good

Ngoguey42:
I'm also going to bench this PR on top of the `try_merge` fix

samoht:
This need to be tested on a real tezos node.

Ngoguey42:
What if, instead of calling `yield` in `replace'` and `Thread.yield` in `merge_with@refill`, we made the merge routine synchronous and _resumable_? We could make sure that the merge computations are spread evenly along the filling of `log_async`.

We could get rid of the threading, and I doubt I would make a difference in performances.

This would make the merge system similar to `Gc.major_slice`.

(@CraigFe Is this what you suggested yesterday when talking about dead locks and #313 ?)

PR #304, opened 2021-04-27
[experimental] Remove yields

pascutto:


samoht:
This need to be tested on a real tezos node.

Issue #296, opened 2021-04-08
Readonly sync takes longer with the `overcommit` option


icristescu:
The `log_async` file is read from scratch at every sync call. This is because, if the generation of `log_async` differs from the generation of the instance then `log_async` is refilled from offset 0 (https://github.com/mirage/index/blob/master/src/index.ml#L245). 

It is especially an issue when using the `overcommit` option, as then the `log_async` can get quite large and a sync from scratch can take a lot of time.

PR #305, opened 2021-04-27
Add test to expose the reload of log_async

icristescu:
as explained in https://github.com/mirage/index/issues/296.


samoht:
Now part of #310

PR #307, opened 2021-04-27
Simplifications on top of #288

samoht:
We don't need to create a temporary file, it's fine to just remove
the initial file while holding an open file descriptor on the resource.

We also can directly remove the files which are not needed after a clear:
only the log file is needed as it contains the current generation number.

### DIFF
--- a/src/unix/index_unix.ml
+++ b/src/unix/index_unix.ml
@@ -157,7 +157,7 @@ module IO : Index.IO = struct
     t.flushed <- t.header;
     Buffer.clear t.buf;
     (* Remove the file current file. This allows a fresh file to be
-       created, before writing the new generation in the temporary file. *)
+       created, before writing the new generation in the old file. *)
     let old = t.raw in
     Unix.unlink t.file;
     (* Open a fresh file. *)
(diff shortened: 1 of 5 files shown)


src/unix/index_unix.ml
```diff
@@ -152,30 +152,25 @@ module IO : Index.IO = struct
...
-  let clear ~generation ?(hook = fun () -> ()) t =
+  let clear ~generation ?(hook = fun () -> ()) ~reopen t =
     t.offset <- Int63.zero;
     t.flushed <- t.header;
     Buffer.clear t.buf;
-    let tmp = t.file ^ "_tmp" in
-    Raw.close t.raw;
-    (* Rename file into a temporary file. This allows a fresh file to be
+    (* Remove the file current file. This allows a fresh file to be
        created, before writing the new generation in the temporary file. *)
```
icristescu, line 160:
```suggestion
       created, before writing the new generation in the old file. *)
```

src/unix/index_unix.ml
```diff
@@ -152,30 +152,25 @@ module IO : Index.IO = struct
...
-    Unix.rename t.file tmp;
-    hook ();
+    let old = t.raw in
```
icristescu, line 161: much simpler, thanks!

icristescu (approved):

PR #308, opened 2021-04-27
use fsync to flush metadata when reoperning a file with new headers

samoht:
Not totally sure about this one: this will probably slow down single writer/multiple instances within the same process (but probably not much as this is not a very common operation) ; but that'll speed up signalling of when a merge is done for other processes opened in read-only mode.


samoht:
Pushed an update. `fsync` is now called after merges to make sure read-only instance are updated correctly. Will merge once the CI is green.

PR #310, opened 2021-04-27
Do not reload log_async on every sync

### LINKED ISSUE #296: Readonly sync takes longer with the `overcommit` option

samoht:
Do not reload log_async on every sync

In general `generation(log_async)` is `generation(log) + 1`. So a read-only instance needs to reload a the `log_async` file only if the diff > 2.

There is also a small race between sync and merge: Merge is performing `clear(log)` then `clear(log_async)` which both bump the generation number. Between these two clears, `generation(log_async) = generation(log)`. So the sync operation also checks if the diff is > 1 and if a `clear(log_async)` has just completed (clear sets `offset` to 0 and unlink the file so noone else could write on it).

Fix #296

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,10 @@
 - Fixed a crash-consistency bug due to a potential flush of an incomplete entry
   to disk. Entries are now flushed as complete strings. (#301)
 
+- Fixed a performance issue for `Index.sync` when there is a blocking merge in
+  progress: the `log_async` file was not cached properly and fully reloaded
+  from disk every time. (#310)
+
 ## Changed
 
 - Specialise `IO.v` to create read-only or read-write instances. (#291)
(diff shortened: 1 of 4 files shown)


icristescu (approved):
The test in https://github.com/mirage/index/pull/305 passes and this changes look good to me!

CraigFe:
It might be good to add an assertion in `IO.v` that the generation number is not `max_int`. I'm somewhat concerned about this value wrapping negative if we violate our invariants.

samoht:
Should be better now - this passes the test and does not use `Int63.max_int` anymore. There's a special case to handle when the last `sync` was done between calls to `clear (log)` and `clear (log_async)`  in a concurrent merge. In that case: 
- `clear (log_async)` will bump the generation number and set offset to zero
- the next sync will not see that the generation number has been bumped and will skip `log_async` update

Issue #284, opened 2021-03-17
Ensure that asynchronous failures propagate cleanly


CraigFe:
As it stands, a variety of unexpected failures in the asynchronous merge thread will cause the process to enter a livelock, since that thread doesn't surrender the merge lock when it fails. We should ensure that it does indeed do that.

PR #312, opened 2021-04-28
index: release the merge lock if a merge raises an exception

CraigFe:
This has no effect on passing tests, but causes failing ones to behave more sensibly in the short term. A more complete solution would fix https://github.com/mirage/index/issues/284 entirely.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -666,9 +666,6 @@ struct
       - [t.log_async] has been created;
       - [t.merge_lock] is acquired before entry, and released immediately after
         this function returns or raises an exception. *)
-  let compare_entry (e : Entry.t) (e' : Entry.t) =
-    compare e.key_hash e'.key_hash
-
   let unsafe_perform_merge ~witness ~filter ~hook t =
     hook `Before;
     let log = Option.get t.log in
@@ -687,7 +684,7 @@ struct
           i + 1)
         log.mem 0
       |> ignore;
-      Array.fast_sort compare_entry b;
+      Array.fast_sort Entry.compare b;
       b
     in
     let fan_size =
--- a/src/index.ml
+++ b/src/index.ml
@@ -689,7 +689,10 @@ struct
     flush_instance ~no_async:() ~with_fsync:true t;
 
     let go () =
-      hook `Before;
+      (try hook `Before
+       with exn ->
+         Semaphore.release t.merge_lock;
+         raise exn);
       let log = Option.get t.log in
       let generation = Int63.succ t.generation in
       let log_array =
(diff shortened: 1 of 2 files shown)


src/index.ml
```diff
@@ -688,7 +688,10 @@ struct
     flush_instance ~no_async:() ~with_fsync:true t;
 
     let go () =
-      hook `Before;
+      (try hook `Before
+       with exn ->
+         Semaphore.release t.merge_lock;
+         raise exn);
```
samoht, line 694: maybe define `hook` to always release the hook? That'll avoid code duplication.

CraigFe: I'll wrap the entire thing in a release to solve https://github.com/mirage/index/issues/284 at the same time.

src/index.ml
```diff
@@ -656,45 +656,25 @@ struct
...
+      writes are not lost.
+
+      The caller must ensure the following:
+
+      - [t.log] has been loaded;
+      - [t.log_async] has been created;
+      - [t.merge_lock] is acquired before entry, and released immediately after
+        this function returns or raises an exception. *)
+  let unsafe_perform_merge =
+    let compare_entry (e : Entry.t) (e' : Entry.t) =
```
samoht, line 669: you can probably move this out of the function (and use `Int.compare` to make sure we don't use the polymorphic compare function) and move the function argument at the "right" place

samoht (approved):
LGTM - there are probably more code simplification that can be done, but that's not critical

src/index.ml
CraigFe: Done in the latest commit.

CraigFe:
Rebased.

src/index.ml
```diff
@@ -657,14 +657,122 @@ struct
...
+        log.mem 0
+      |> ignore;
+      Array.fast_sort Entry.compare b;
+      b
+    in
+    let fan_size =
+      match t.index with
+      | None -> Tbl.length log.mem
+      | Some index ->
+          (Int63.to_int (IO.offset index.io) / Entry.encoded_size)
```
samoht, line 694: This is broken in 32 bit but we probably do not care :-)

src/index.ml
```diff
@@ -657,14 +657,122 @@ struct
...
+        merge_path
+    in
+    let merge_result : [ `Index_io of IO.t | `Aborted ] =
+      match t.index with
+      | None -> (
+          match check_pending_cancel t with
+          | `Abort -> `Aborted
+          | `Continue ->
+              let io =
+                IO.v ~fresh:true ~generation ~fan_size:Int63.zero
```
samoht, line 713: Isn't this `IO.clear` ?

samoht:
LGTM

Issue #311, opened 2021-04-28
`Index.try_merge` doesn't check for an ongoing merge


CraigFe:
The `Index.try_merge` function will run a merge iff the log is full, regardless of whether there is currently a merge ongoing. This isn't technically a bug, since it's implied by the docstrings of `merge` + `try_merge` taken together:

```ocaml
  val merge : t -> unit
  (** [merge t] forces a merge for [t].

      If there is no merge running, this operation is non-blocking, i.e. it
      returns immediately, with the merge running concurrently.

      If a merge is running already, this operation blocks until the previous
      merge is complete. It then launches a merge (which runs concurrently) and
      returns. *)

  val try_merge : t -> unit
  (** [try_merge] is like {!merge} but is a no-op if the number of entries in
      the write-ahead log is smaller than [log_size]. *)
```

... but I think this behaviour is unexpected regardless. My intuition is that `try_merge` is intended as a mechanism for the user to hint that now is a good time to launch a merge (whereas the current semantics amount to "Ensure that the log is not full, blocking if necessary".)

This behaviour should be simple to reproduce from within Index by launching an async merge and then attempting `try_merge` concurrently (using hooks).

CraigFe:
I came back to #313 today and realised I don't really understand the intended use-case of `try_merge`. My understanding is that it was added as a "hint" mechanism for triggering merges just after having completed a batch of writes, but I think this leads to conflicting requirements:

- we want `try_merge` to be a no-op when the log is not full, otherwise hinting too often will ruin performance as we fail to use the available memory;
- we want `try_merge` to be a no-op when there is already a merge taking place, as stacking merges blocks the main thread unnecessarily;
- _but_, the only scenario where the log is full _and_ there's not a merge ongoing is when using ``~throttle:`Overcommit_memory`` and overfilling the log during a prior merge. In this case, the merge will be triggered on the next `replace` anyway, so `try_merge` is only able to start the merge slightly earlier than would otherwise be the case.

@icristescu, @samoht: do I understand the situation correctly? how would we like `try_merge` to behave?

icristescu:
I think we added this for the layered store:  we have several indexes and we don't want merges to be triggered at the same time for all of them. And so we need some control over when the merges are called. 

[...]

PR #313, opened 2021-04-28
index: add a reproduction for try_merge blocking issue

CraigFe:
See https://github.com/mirage/index/issues/311. Based on https://github.com/mirage/index/pull/312 since otherwise the failing test just deadlocks.


test/unix/main.ml
```diff
@@ -853,6 +853,32 @@ module Throttle = struct
...
+            "Lock not previously held" true not_previously_held
+      | _ -> ()
+    in
+    for _ = 1 to log_size do
+      ignore (add_binding rw : binding)
+    done;
+    Log.app (fun m -> m "Triggering an implicit merge");
+    let _, merge_result = add_overflow_binding ~hook rw in
+    Log.app (fun m -> m "Running `try_merge` while a merge is ongoing");
+    Index.try_merge rw;
```
icristescu, line 875: shouldn't `try_merge` use the hook as well?

CraigFe: Indeed, this mechanism for synchronising with the merge thread is broken. I'm now remembering why we didn't write tests this way to begin with... I'll propose something else as part of a different change.

PR #316, opened 2021-04-28
Raise exceptions on failed Int63.to_int coercions

CraigFe:
Fixes the bug spotted here: https://github.com/mirage/index/pull/312#discussion_r622221691 (not that anyone's likely to have run into it). I think `Int63.to_int` is a little too tempting from an API perspective, so I've shadowed it and provided more clearly named checked and unchecked alternatives.


CraigFe:
Rebased.

PR #317, opened 2021-04-28
Simplify clear again

samoht:
/cc @icristescu


samoht:
As discussed with @icristescu I have removed the unrelated `assert` false in `sync_instance`

Issue #314, opened 2021-04-28
Index links against `mtime.clock.os`


CraigFe:
The core of Index is currently using `Mtime_clock` to provide timings of various operations in the merge. This adds a Unix dependency that should probably be inside `irmin.unix` instead (and passed to Index as a functor parameter as part of the platform dependencies).

PR #321, opened 2021-04-30
Refactor use of monotonic time in Index

CraigFe:
- avoid using `Sys.time` to measure durations;
- avoid dependency on `mtime.clock.os` in Index core by functorising over a time source.

... and does some minor refactoring to extract `Logs` boilerplate and cleanup.

Fix https://github.com/mirage/index/issues/314.

### DIFF
--- a/bench/bench.ml
+++ b/bench/bench.ml
@@ -60,15 +60,24 @@ end
 
 let with_stats f =
   Stats.reset_stats ();
-  let t0 = Sys.time () in
-  let _ = f () in
-  let t1 = Sys.time () -. t0 in
+  let _, duration = Common.with_timer f in
   let stats = Stats.get () in
-  (t1, stats)
+  (duration, stats)
+
+module Mtime = struct
+  include Mtime
+
+  module Span = struct
+    include Span
+
+    let to_yojson s = `String (Fmt.to_to_string pp s)
+    let div a b = of_uint64_ns (Int64.div (to_uint64_ns a) (Int64.of_int b))
+  end
+end
 
 module Benchmark = struct
   type result = {
-    time : float;
+    time : Mtime.Span.t;
     ops_per_sec : float;
     mbs_per_sec : float;
     read_amplification_calls : float;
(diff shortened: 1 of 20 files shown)


src/index.ml
```diff
@@ -933,11 +938,12 @@ struct
     ignore (replace' ?hook:None ?overcommit t key value : _ async option)
 
   let replace_with_timer ?sampling_interval t key value =
-    if sampling_interval <> None then Stats.start_replace ();
-    replace t key value;
     match sampling_interval with
-    | None -> ()
-    | Some sampling_interval -> Stats.end_replace ~sampling_interval
+    | None -> replace t key value (* XXX(craigfe): why not log duration here? *)
```
icristescu, line 942: it's just a convention we made for when `sampling_interval` is not set https://github.com/mirage/index/blob/master/src/index_intf.ml#L255

icristescu (approved):
Looks great, thanks

bench/bench.ml
```diff
@@ -79,10 +88,11 @@ module Benchmark = struct
     merges_duration : float; [@key "merges_duration_us"]
     nb_merges : int;
   }
-  [@@deriving yojson]
+  [@@deriving to_yojson]
```
samoht, line 91: unrelated but can we use `repr` here?

src/import.ml
```diff
@@ -1,3 +1,42 @@
...
+module Logs = struct
+  let default_reporter ?(prefix = "") () =
+    let report src level ~over k msgf =
+      let k _ =
+        over ();
+        k ()
+      in
+      let ppf = match level with Logs.App -> Fmt.stdout | _ -> Fmt.stderr in
+      let with_stamp h _tags k fmt =
+        let dt = Unix.gettimeofday () in
```
samoht, line 14: Not sure we want to use `gettimeoday` here

src/index.ml
CraigFe: OK, thanks!

src/import.ml
CraigFe: I've pushed a commit that switches it to `Mtime` instead. Could also use `Ptime`, but I'm not sure which is more applicable here.

bench/bench.ml
CraigFe: Not without defining a generic Yojson codec in Repr, which we should probably do at some point (we're currently using `jsonm` only).

PR #322, opened 2021-05-05
Release the prefetch warning threshold

pascutto:
The current setting is probably a bit too defensive, as unevenly distributed keys happens quite often in practice.

### DIFF
--- a/src/io_array.ml
+++ b/src/io_array.ml
@@ -79,8 +79,8 @@ module Make (IO : Io.S) (Elt : ELT) :
     (* The prefetched area should not exceed 4096 in most cases, thanks to the
        fan out. However, if the hash function is not well distributed, some
        exceptions might happen where the prefetched area actually exceeds 4096.
-       As long as this excess is reasonable (x2), we still want to prefetch. *)
-    2 * 4096
+       As long as this excess is reasonable (x8), we still want to prefetch. *)
+    8 * 4096
 
   let buf = Bytes.create max_buffer_size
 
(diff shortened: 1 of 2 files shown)


src/io_array.ml
```diff
@@ -79,8 +79,8 @@ module Make (IO : Io.S) (Elt : ELT) :
     (* The prefetched area should not exceed 4096 in most cases, thanks to the
        fan out. However, if the hash function is not well distributed, some
        exceptions might happen where the prefetched area actually exceeds 4096.
-       As long as this excess is reasonable (x2), we still want to prefetch. *)
-    2 * 4096
+       As long as this excess is reasonable (x8), we still want to prefetch. *)
```
samoht, line 82: does it force all reads to read 8 full pages?

pascutto: It should not, the amount it reads depends on what the fan-out says: https://github.com/mirage/index/blob/master/src/io_array.ml#L88-L91, it's just an upper bound.

samoht: Are we loading this buffer on every `find`?

CraigFe:
(Even though this is a tiny change, I think it could still do with a CHANGES entry since it impacts Tezos.)

icristescu (approved):
Thanks!

src/io_array.ml
pascutto: Not sure what you mean by "load", but every `find` that has to look in the `data` file reads the disk to (partially) fill it, then (interpolation-)searches into it indeed.

pascutto:
> (Even though this is a tiny change, I think it could still do with a CHANGES entry since it impacts Tezos.)

Done

PR #324, opened 2021-05-07
Fan warning and bench updated with a log reporter

mattiasdrp:
Fan warning is invoked when the fan-out is double in size

The log reporter is quiet by default when using `make bench` and shall
be invoked with `--verbosity {error | warning | info | debug}`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,8 @@
 
 - Added benchmarks that replay a trace of index operations. (#300)
 
+- Log reporter for the benches
+
 # 1.3.0 (2021-01-05)
 
 ## Added
--- a/bench/bench.ml
+++ b/bench/bench.ml
@@ -149,6 +149,8 @@ let bindings_pool = ref [||]
 let absent_bindings_pool = ref [||]
 let sorted_bindings_pool = ref [||]
 
+module Index_lib = Index
+
 module Index = struct
   module Index =
     Index_unix.Private.Make (Context.Key) (Context.Value) (Index.Cache.Noop)
@@ -639,7 +641,8 @@ let minimal_flag =
 let cmd =
   let doc = "Run all the benchmarks." in
   ( Term.(
-      const run
+      const (fun () -> run)
+      $ Index_lib.Private.Logs.setup_term (module Mtime_clock)
       $ name_filter
       $ data_dir
       $ output
--- a/src/index.ml
+++ b/src/index.ml
@@ -709,15 +709,26 @@ struct
       Array.fast_sort Entry.compare b;
       b
     in
-    let fan_size =
+    let fan_size, old_fan_nb =
       match t.index with
-      | None -> Tbl.length log.mem
+      | None -> (Tbl.length log.mem, None)
       | Some index ->
-          Int63.(to_int_exn (IO.offset index.io / entry_sizeL))
-          + Array.length log_array
+          ( Int63.(to_int_exn (IO.offset index.io / entry_sizeL))
+            + Array.length log_array,
+            Some (Fan.nb_fans index.fan_out) )
     in
     let fan_out =
       Fan.v ~hash_size:K.hash_size ~entry_size:Entry.encoded_size fan_size
+    in
+    let () =
+      match old_fan_nb with
+      | Some on ->
+          let new_fan_nb = Fan.nb_fans fan_out in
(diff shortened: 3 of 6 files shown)


bench/bench.ml
```diff
@@ -640,6 +642,7 @@ let cmd =
   let doc = "Run all the benchmarks." in
   ( Term.(
       const run
+      $ Index_lib.Private.Logs.setup_term (module Mtime_clock)
```
CraigFe, line 645:
```suggestion
      const (fun () -> run)
      $ Index_lib.Private.Logs.setup_term (module Mtime_clock)
```

We usually try to keep the non-Cmdliner API independent of the idiosyncrasies required by the Cmdliner one.

src/fan.mli
```diff
@@ -22,9 +22,11 @@ type 'a t
...
 (** The equality function for fanouts. *)
 
-val v : hash_size:int -> entry_size:int -> int -> [ `Write ] t
-(** [v ~hash_size ~entry_size n] creates a fan_out for an index with [hash_size]
-    and [entry_size], containing [n] elements. *)
+val v : hash_size:int -> entry_size:int -> int -> int option -> [ `Write ] t
+(** [v ~hash_size ~entry_size n os] creates a fan_out for an index with
+    [hash_size] and [entry_size], containing [n] elements. If the fan is created
+    while another one was already present, a warning is issued if the new fan is
+    twice the previous one *)
```
CraigFe, line 29: The fan-out implementation itself is agnostic to the existence of a previous fan-out, so I think this logging should be done externally. i.e. build the new fan out, query `Fan.nb_fans`, then log the size difference as part of the merge logic.

src/fan.ml
```diff
@@ -32,13 +32,20 @@ let equal t t' =
...
   let entries_per_page = 4096. /. entry_sizef in
   let raw_nb_fans = float_of_int n /. entries_per_page in
   let size = max 0 (int_of_float (ceil (log2 raw_nb_fans))) in
   let nb_fans = 1 lsl size in
   let shift = hash_size - size in
+  let () =
+    match old_fan_nb with
+    | Some on when nb_fans > on ->
+        Log.info (fun m ->
+            m "number of fans has double (from %d to %d)" on nb_fans)
```
CraigFe, line 46:
```suggestion
            m "the number of fan-out entries has changed: %d to %d" on nb_fans)
```

It's technically possible for it to change by any power of two (increasing by more than double if the log size is increased between runs, shrinking if `Index.filter` is used to drop elements).

CHANGES.md
```diff
@@ -34,6 +34,10 @@
 
 - Added benchmarks that replay a trace of index operations. (#300)
 
+- Warning when the fan-out size doubles
+
```
CraigFe, line 38: Since this is `info` now; we probably don't need a CHANGES for it.

mattiasdrp:
Thanks for the review, pushed the changes :-)

src/fan.mli
CraigFe: (API and comment change  now need to go too.)

src/index.ml
```diff
@@ -709,15 +709,27 @@ struct
...
     in
     let fan_out =
       Fan.v ~hash_size:K.hash_size ~entry_size:Entry.encoded_size fan_size
+        old_fan_nb
+    in
+    let () =
+      match old_fan_nb with
+      | Some on ->
+          let new_fan_nb = Fan.nb_fans fan_out in
+          if new_fan_nb > on then
```
CraigFe, line 728:
```suggestion
          if new_fan_nb <> on then
```

test/fuzz/fan/main.ml
```diff
@@ -11,7 +11,7 @@ let hash = map [ bytes ] Hashtbl.hash
 let hash_list = map [ list hash ] (List.sort compare)
 
 let empty_fan_with_size =
-  map [ bounded_int ] (fun n -> (Fan.v ~hash_size ~entry_size n, n))
+  map [ bounded_int ] (fun n -> (Fan.v ~hash_size ~entry_size n None, n))
```
CraigFe, line 14: (Should be unnecessary now.)

CraigFe:
CI is still unhappy here.

mattiasdrp:
Commits squashed. Ready to be merged.

PR #328, opened 2021-06-06
Avoid using an intermediate scratch buffer in the merge

CraigFe:
The merge process loads entries from the old data file into a buffer before writing them out to the new one. Previously, individual entries were then copied to a second buffer before writing, but this isn't necessary: the IO can simply read substrings from the first one directly.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -575,9 +575,9 @@ struct
 
   (** Appends the entry encoded in [buf] into [dst_io] and registers it in
       [fan_out] with hash [hash]. *)
-  let append_buf_fanout fan_out hash buf_str dst_io =
+  let append_substring_fanout fan_out hash dst_io buf ~off ~len =
     Fan.update fan_out hash (IO.offset dst_io);
-    IO.append dst_io buf_str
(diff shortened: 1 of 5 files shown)


src/index.ml
```diff
@@ -654,11 +653,10 @@ struct
...
               not K.(equal log_key index_key))
               && filter index_key buf_offset
-            then (
-              Bytes.blit buf buf_offset buf_str 0 Entry.encoded_size;
-              append_buf_fanout fan_out index_key_hash
-                (Bytes.unsafe_to_string buf_str)
-                dst_io);
+            then
+              append_substring_fanout fan_out index_key_hash dst_io
+                (Bytes.unsafe_to_string buf)
```
mattiasdrp, line 658: I wanted to be sure that using `Bytes.unsafe_to_string` didn't create extra allocations and it doesn't so everything is ok for me! 

Great PR, let's merge it :-)

mattiasdrp (approved):

Ngoguey42 (approved):
LGTM

CraigFe:
Thanks for the reviews :slightly_smiling_face:

PR #329, opened 2021-06-06
Attempt to recover from a broken IO header invariant

CraigFe:
In the event that the `log_async` headers change unexpectedly when attempting a sync, we can attempt a recovery by reloading the entire `log` from scratch (as this may contain new updates).

In principle, this case should never occur, but the previous `assert false` has been triggered via Tezos at least once (https://gitlab.com/tezos/tezos/-/issues/1403), and the failing node was able to recover on restart.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,9 @@
 - Added fsync after `Index.clear` to signal more quickly to read-only instances
   than something has changed in the file (#308)
 
+- Attempt to recover from `log_async` invariant violations during an explicit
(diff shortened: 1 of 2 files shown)


CraigFe:
Since I don't know how this case arises, I'm not sure how to test it. A careful review would be appreciated.

Ngoguey42 (commented):
Your changes are straightforward and look good to me, but I'm no expert on Index.

icristescu (approved):
I don't understand how this occurs either, but your changes look good to me, LGTM.

PR #336, opened 2021-07-12
Avoid flaky usage of [@tailcall]

CraigFe:
As of `4.13.0~alpha`, function applications on the left-hand-side of an application of `( @@ )` are rejected as not tailcalls.


CraigFe:
Upstream issue: https://github.com/ocaml/ocaml/issues/10507.

samoht:
Not sure why the CI is all red here.

CraigFe:
Potentially just because I hadn't rebased -- we'll see.

CraigFe:
Looks like that was it. Merging, even though the need for this may disappear by the time 4.13 is actually out.

PR #337, opened 2021-07-12
test: add more tests for re-opening of stores

samoht:


samoht:
Added more tests for log/log_async transfert.

samoht:
Now folded into #339

PR #338, opened 2021-07-13
Add an fsync after flushing the log file on startup

samoht:
On startup, we transfer entries from log_async to log. Make sure to call fsync here so that persisted entries in log_async continue to be persisted in log.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,9 @@
   continue to run and will need to recover from the crash before doing a new
   merge. This fixes a critical issue which might cause data loss (#339)
 
+- Make sure that no entries can disappear for read-only instances during
+  log_async recovery (#338)
+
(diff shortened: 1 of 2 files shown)


samoht:
Might be related to https://github.com/mirage/irmin/issues/1476 but we need more clues

samoht:
Rebased to pick up the CI fixes.

CHANGES.md
```diff
@@ -4,6 +4,9 @@
 
 - Proper cleaning of merge file descriptors when aborting a merge (#326)
 
+- Make sure that no entries can disapear for read-only instances during
```
CraigFe, line 7:
```suggestion
- Make sure that no entries can disappear for read-only instances during
```

CraigFe (approved):
LGTM.

PR #339, opened 2021-07-13
Recover crashed merges

samoht:


test/unix/main.ml
```diff
@@ -623,11 +623,151 @@ end
...
+  let force_merge rw =
+    let thread = Index.try_merge_aux ~force:true rw in
+    Index.await thread |> check_completed;
+    check_logs "force_merge" rw (Some 0) None
+
   let close_reopen_rw () =
-    let* Context.{ rw; tbl; clone; _ } = Context.with_full_index () in
-    Index.close rw;
-    let w = clone ~readonly:false () in
-    check_equivalence w tbl
```
CraigFe, line 630: Perhaps there's value in keeping a shorter test that doesn't trigger a crash? (And adding the other ones too.)

test/unix/main.ml
```diff
@@ -623,11 +623,151 @@ end
...
+    Hashtbl.add tbl k1 v1;
+    check_equivalence rw tbl;
+
+    (* Add k2 in log_async, as the file is present from the previous
+       crash. *)
+    Index.replace rw k2 v2;
+    Index.flush rw;
+    check_logs __LOC__ rw (Some 0) (Some 1);
+
+    (* k3 should be there. *)
```
CraigFe, line 749:
```suggestion
    (* k2 should be there. *)
```

test/unix/main.ml
```diff
@@ -623,11 +623,151 @@ end
...
+    Hashtbl.add tbl k2 v2;
+    check_equivalence rw tbl;
+
+    (* merge should merge pre-existing log_async entries  *)
+    Index.try_merge_aux ~force:true ~hook:merge_hook rw |> wait;
+    check_logs __LOC__ rw (Some 0) (Some 0);
+    check_equivalence rw tbl;
+
+    (* Add k3 in log_async as it already exists *)
+    Index.replace ~overcommit:true rw k3 v3;
```
CraigFe, line 759:
```suggestion
    Index.replace rw k3 v3;
```

(Not necessary, I think.)

test/unix/main.ml
CraigFe, line 749:
```suggestion
    (* k2 should be there. *)
```

test/unix/main.ml
CraigFe, line 759:
```suggestion
    Index.replace ~overcommit:true rw k3 v3;
```

(Not necessary, I think.)

test/unix/main.ml
```diff
@@ -623,11 +623,151 @@ end
...
+      I.Private.Hook.v @@ function
+      | `After_clear ->
+          Semaphore.release merge;
+          raise Stop
+      | _ -> ()
+    in
+    let* Context.{ rw; tbl; _ } = Context.with_full_index ~log_size:2 () in
+    let k1, v1 = (Key.v (), Value.v ()) in
+    let k2, v2 = (Key.v (), Value.v ()) in
+    let k3, v3 = (Key.v (), Value.v ()) in
```
CraigFe, line 718: (Or use `let k1, v1 = Index.replace_random rw in` below.)

CraigFe (commented):
Looks good. Only had one comment about correctness of the fix (related to the case `t.log = None`), and the various nits can be saved for later if necessary.

src/index.ml
```diff
@@ -438,6 +448,23 @@ struct
 
   (** {1 Index creation} *)
 
+  let transfer_log_async_to_log ~root ~generation ~log ~log_async =
+    let entries = Int63.div (IO.offset log_async) entry_sizeL in
+    Log.debug (fun l ->
+        l "[%s] log_async file detected. Loading %a entries"
+          (Filename.basename root) Int63.pp entries);
+    Option.iter
```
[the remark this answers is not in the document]
samoht: t.log is always supposed to be `Some _`. We should probably refactor this in the codebase to make this clearer.

src/index.ml
```diff
@@ -477,29 +504,15 @@ struct
...
-            let append_io = IO.append log.io in
-            iter_io
-              (fun e ->
-                Tbl.replace log.mem e.key e.value;
-                Entry.encode e append_io)
-              io;
-            IO.flush log.io;
-            IO.clear ~generation ~reopen:false io)
-          log);
+    (if (not readonly) && IO.exists log_async_path then
```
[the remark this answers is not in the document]
samoht: indeed, the also need to be fixed; have opened #340 to track this

src/index.ml
```diff
@@ -477,29 +504,15 @@ struct
...
+     let io =
+       IO.v ~flush_callback ~fresh ~generation:Int63.zero ~fan_size:Int63.zero
+         log_async_path
+     in
+     (* If we are not in fresh mode, we move the contents of log_async to
+        log. *)
+     if not fresh then
+       transfer_log_async_to_log ~root ~generation ~log ~log_async:io);
```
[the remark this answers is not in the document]
samoht: I could also move the `not fresh` in the initial `if`. Will simplify this.

src/index.ml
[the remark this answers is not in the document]
samoht: `t.log = None` can only occurs in read-only mode. Will add a comment for this, and if we ever want to split the rw/ro state we could fix it later.

src/index.ml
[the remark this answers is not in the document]
CraigFe: I remember thinking hard merging these two: in the `fresh:true` case it's still necessary to `IO.v` in order to wipe the `log_async`.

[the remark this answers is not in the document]
samoht: ha yes good point! but in that case I am not sure why we use generation=0 here. Shouldn't it be using the same generation number as `log`? Also, we never close that `fd`!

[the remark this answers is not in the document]
icristescu: For the generation number, we could set it to the log's generation to be consistent but it does not change anything - the values of log_async are in log until the next merge starts; once the next merge starts, the the log_async's generation is rewritten https://github.com/mirage/index/blob/master/src/index.ml#L830

test/unix/main.ml
```diff
@@ -623,11 +623,151 @@ end
...
+      | _ -> Alcotest.fail "the merge thread should have been killed"
+    in
+
+    (* empty log and log_async *)
+    force_merge rw;
+
+    (* Add k1, start a merge and crash just after the index is
+       renamed *)
+    Index.replace rw k1 v1;
+    Index.flush rw;
```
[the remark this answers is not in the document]
samoht: it is now :-) I've modified `Index.log` and `Index.log_async` to fail if the in-memory representation is different from the on-disk contents

test/unix/main.ml
samoht: I've re-added this simple test

samoht:
Rebased and I've tried to fix the issues identified during the review.

samoht:
Rebased to pick the CI fixes.

src/index.ml
```diff
@@ -474,32 +514,24 @@ struct
...
+      let log_async_path = Layout.log_async ~root in
+      (* - If we are in readonly mode, the log_async will be read
+           during sync_log so there is no need to do it here. *)
+      if (not readonly) && (not fresh) && IO.exists log_async_path then
+        let io =
+          IO.v ~flush_callback ~fresh ~generation ~fan_size:Int63.zero
+            log_async_path
+        in
+        (* in fresh mode, we need to wipe the existing [log_async] file. *)
+        if fresh then IO.clear ~generation ~reopen:false io
```
Ngoguey42, line 528: This `fresh = true` branch is unreachable because of the `(not fresh)` test above

Ngoguey42 (commented):
I think we should also log an error when the merge thread crashes. Maybe from [there](https://github.com/mirage/index/blob/415a85/src/unix/index_unix.ml#L357-L359)

src/index.ml
samoht: woops I forgot to remove it - well spotted!

src/index.ml
```diff
@@ -474,32 +514,24 @@ struct
...
-              io;
-            IO.flush log.io;
-            IO.clear ~generation ~reopen:false io)
-          log);
+    (* recover [log_async] entries *)
+    let () =
+      let log_async_path = Layout.log_async ~root in
+      (* - If we are in readonly mode, the log_async will be read
+           during sync_log so there is no need to do it here. *)
+      if (not readonly) && (not fresh) && IO.exists log_async_path then
```
samoht, line 522:
```suggestion
      if (not readonly) && IO.exists log_async_path then
```

src/index.ml
```diff
@@ -159,6 +159,27 @@ struct
...
+
+  let io_to_list msg io =
+    let x = to_list io.mem in
+    let y = ref [] in
+    iter_io (fun (e : Entry.t) -> y := (e.key, e.value) :: !y) io.io;
+    if List.length x <> List.length !y then (
+      let pp_entry ppf (k, _) = pp_key ppf k in
+      let pp = Fmt.Dump.list pp_entry in
+      Fmt.epr "consistency error in %s:\nmem : %a\ndisk: %a\n%!" msg pp x pp !y;
+      assert false);
```
samoht, line 172: After reading this again: probably not the best API here - it works only if the io instance has been flushed to disk before calling the function. I'll try to make it clearer.

samoht:
Updated with a better API for `Index.log` and `Index.log_async`: they are now the ones responsible for calling `flush` (and so this simplifies the tests as well, as requested by @CraigFe )

src/index.ml
```diff
@@ -438,6 +463,24 @@ struct
...
+      (fun log ->
+        let append_io = IO.append log.io in
+        iter_io
+          (fun e ->
+            Tbl.replace log.mem e.key e.value;
+            Entry.encode e append_io)
+          log_async;
+        IO.flush log.io;
+        IO.clear ~generation ~reopen:false log_async)
+      log
```
CraigFe, line 482:
```suggestion
    match log with
    | None -> assert false
    | Some log -> (
        let append_io = IO.append log.io in
        iter_io
          (fun e ->
            Tbl.replace log.mem e.key e.value;
            Entry.encode e append_io)
          log_async;
        IO.flush log.io;
        IO.clear ~generation ~reopen:false log_async)
```

src/index.ml
```diff
@@ -474,32 +518,24 @@ struct
...
-            let append_io = IO.append log.io in
-            iter_io
-              (fun e ->
-                Tbl.replace log.mem e.key e.value;
-                Entry.encode e append_io)
-              io;
-            IO.flush log.io;
-            IO.clear ~generation ~reopen:false io)
-          log);
+    (* recover [log_async] entries *)
```
CraigFe, line 521:
```suggestion
    (* load the [log_async] file *)
```

(It's possible to `close ~immediately:()`, so this isn't necessarily crash recovery.)

CraigFe (approved):

PR #353, opened 2021-09-06
Separate raw stats from index stats

icristescu:
This PR has two objectives:
- separate the raw stats from the index one - different libraries than index can use the `raw.ml` (for instance cactus), and this allows to have the IO stats even if `index` is not used;
- different stats for `merge` and the rest of the operations - now stats are reported by file name and so we can report IO stats only due to the merge


src/platform.ml
```diff
@@ -58,9 +58,35 @@ module type THREAD = sig
...
+
+  val get : unit -> t
+  val get_by_file : string -> t
+  val get_all : unit -> (string * t) list
+  val reset : unit -> unit
+  val bytes_read : unit -> int
+  val bytes_written : unit -> int
+  val nb_reads : unit -> int
+  val nb_writes : unit -> int
+end
```
Ngoguey42, line 84: Could you add documentation here to explain the need for 2 separate interfaces?

src/platform.ml
```diff
@@ -58,9 +58,35 @@ module type THREAD = sig
...
+  val reset : t -> unit
+end
+
+module type IO_STATS = sig
+  include RAW_STATS
+
+  val get : unit -> t
+  val get_by_file : string -> t
+  val get_all : unit -> (string * t) list
+  val reset : unit -> unit
```
Ngoguey42, line 79: You could rename that one to `reset_all` to avoid shadowing `val fresh_stats : unit -> t`

Ngoguey42 (approved):

src/platform.ml
icristescu: Added some [doc](https://github.com/mirage/index/pull/353/files#diff-f5a28a7f00acf9e80c4d78a180ab01f0cb888ae685b65347a9108d5e4faa420fR62), let me know if its clear.

Ngoguey42: 👍

Issue #354, opened 2021-10-03
Reduce the memory usage of the write-ahead log


CraigFe:
### Context

The total IO done by an index writer is roughly proportional to the maximum size of the log file, since the store must be entirely rewritten once per `log_size` replace operations. As a result, the in-memory mirror of the log file tends to dominate the overall memory usage (i.e. since it's worth making `log_size` as large as possible). When used in `irmin-pack`, this data-structure expands out as follows:

```ocaml
  (key, value) Stdlib.Hashtbl.t
= (string<32>, (int63 * int * char)) Stdlib.Hashtbl.t
= (string<32>, (int63 * int * char)) list2 array
```

When `Sys.word_size = 64`, this makes for approximately **14.5 words per binding**, broken down into:

- 6 word string (header + 4 words data + null padding)
- 4 word triple (header + 3 fields)
- 4 word list node (header + key pointer + value pointer + tail pointer)
- ~0.5 word bucket slot (in the best case, there are ~2 bindings per bucket)

In the worst case, there are both reader and writer instances and merges are stacking, meaning there are 4 such hashtables (`log` and `log_async` for each of the reader and writer) and the writer also has an array of sorted entries (5 words per binding). When `log_size = 500_000`, this sums to ~250MB. Here's two (conflicting) optimisations we could make:

### Suggestion 1: functorise over a user-defined hashtable implementations

We could provide a functor that allows the user to provide their own `Hashtbl` implementation specialised to their particular `key` and `value` types. In the specific case of `irmin-pack`, we could do the following:

- keep `string` keys in an arena (-2 words)
- use a small-size optimised list for hashtable buckets (-1.5 words)
- pack the value triple into a pair of `int63`s (`offset` and `kind_then_length`) and keep them as immediates in the hashtable (-3 words)

This makes for **8 words per binding**. By itself, this isn't much of a change since the entries must be unpacked from their compact representation to sort them in an array before the merge. The simple solution here is just to pick the hashtable bucket using the upper bits of `key_hash`, so that the hashtable is already almost sorted in the correct order (only the bindings within a particular bucket are relatively out-of-order). This means we can expose a `to_sorted_seq` that takes O(1) additional memory.

When `log_size = 500_000`, this works out to ~128MB in the worst case. (Half what we had before.)

### Suggestion 2: keep only log offsets in memory

[...]

PR #355, opened 2021-10-04
Reduce the memory usage of the write-ahead log

CraigFe:
This implements "Suggestion 2" as described in https://github.com/mirage/index/issues/354, without adding an LRU.

Fix https://github.com/mirage/index/issues/354.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -73,10 +73,15 @@ struct
     include Stats
   end
 
-  module Tbl = Hashtbl.Make (K)
-
   module IO = struct
     include Io.Extend (IO)
+
+    let iter_keys ?min f =
+      let page_size = Int63.(mul entry_sizeL (of_int 1_000)) in
+      iter ~page_size ?min (fun ~off ~buf ~buf_off ->
+          let key, _ = Entry.decode_key buf buf_off in
+          f off key;
+          Entry.encoded_size)
 
     let iter ?min ?max f =
       let page_size = Int63.(mul entry_sizeL (of_int 1_000)) in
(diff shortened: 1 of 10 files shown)


src/log_file.ml
```diff
@@ -0,0 +1,207 @@
...
+  let reload t =
+    clear_memory t;
+    sync_entries ~min:Int63.zero t
+
+  let create io =
+    let cardinal = Int63.(to_int_exn (IO.offset io / Entry.encoded_sizeL)) in
+    let bucket_count_log2, bucket_count =
+      let rec aux n_log2 n =
+        if n >= cardinal then (n_log2, n)
+        else if n * 2 > Sys.max_array_length then (n_log2, n)
```
icristescu, line 165: since we are guarding against this case here, maybe we should do it when resizing too? (though it seems likely to reach such large arrays)

src/log_file.mli
```diff
@@ -0,0 +1,63 @@
...
+open! Import
+
+module Make (IO : Io.S) (Key : Data.Key) (Value : Data.Value) : sig
+  module Entry : Data.Entry.S with type key := Key.t and type value := Value.t
+
+  type t
+  type key := Key.t
+  type value := Value.t
+
+  val create : IO.t -> t
```
icristescu, line 26: maybe add a comment here that it also syncs the in-memory data with the disk? since we used to do this in two  steps before (the same for `replace`)

src/unix/index_unix.ml
```diff
@@ -85,8 +85,31 @@ module IO : Index.Platform.IO = struct
...
+      let bytes_from_disk = max 0 (Int63.to_int (t.flushed -- off)) in
+      let bytes_from_buffer = len - bytes_from_disk in
+      let () =
+        if bytes_from_disk > 0 then
+          let read = Raw.unsafe_read t.raw ~off ~len:bytes_from_disk buf in
+          assert (read = bytes_from_disk)
+      in
+      let () =
+        let src_off = max 0 (Int63.to_int (off -- t.flushed)) in
+        let len = min (Buffer.length t.buf - src_off) bytes_from_buffer in
```
icristescu, line 109: tricky function :) 
`len` is diferrent from `bytes_from_buffer`  only when we read more than we wrote? maybe we should fail then?

src/small_list.ml
```diff
@@ -0,0 +1,192 @@
...
+ *)
+
+type 'a t =
+  | Tuple0
+  | Tuple1 of 'a
+  | Tuple2 of 'a * 'a
+  | Tuple3 of 'a * 'a * 'a
+  | Tuple4 of 'a * 'a * 'a * 'a
+  | Tuple5 of 'a * 'a * 'a * 'a * 'a
+  | Tuple6 of 'a * 'a * 'a * 'a * 'a * 'a
```
icristescu, line 24: maybe this is obvious, but why up to 6?

icristescu (approved):
A lot of nice optimizations, thanks!

src/unix/index_unix.ml
CraigFe: > tricky function :)

Yeah, I needed pen and paper for this one :sweat_smile:

> `len` is diferrent from `bytes_from_buffer` only when we read more than we wrote?

The problem here is that a writer may request more bytes than exist (in particular, [here](https://github.com/mirage/index/blob/201c5a66871c41467af0a360dc0ad5d2c28ac3b1/src/index.ml#L664)), in which case we just read until the end of the write buffer (`Buffer.length t.buf - src_off`). I've refactored this a bit to hopefully make it clearer what's going on.

(At some point, we should have a library with buffered reader / writer abstractions so we don't have to re-invent the wheel here.)

src/small_list.ml
CraigFe: Not obvious at all; I just got bored after 6 :wink: In practice, sizes above about 7 don't really happen with the hash functions we're using, so it's not worth going any further. If we used `Obj` rather than a type-safe implementation, we could push the threshold much higher & also improve the speed.

As it happens, @pascutto has a WIP patch to use open-addressing in the hashtable, which would avoid the need for `small_list.ml` entirely. For now, this is probably good enough.

src/unix/index_unix.ml
icristescu: I see, thanks!

src/log_file.ml
```diff
@@ -0,0 +1,212 @@
...
+    in
+    let hashset = Array.make bucket_count Small_list.empty in
+    let t =
+      {
+        io;
+        append_io = IO.append io;
+        hashset;
+        bucket_count_log2;
+        scratch_buf = Bytes.create Entry.encoded_size;
+        cardinal = 0;
```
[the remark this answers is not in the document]
CraigFe: Indeed, good spot. (Artefact from an earlier implementation that didn't load on `create`.)

src/log_file.ml
```diff
@@ -0,0 +1,212 @@
...
+          f off key;
+          Entry.encoded_size)
+  end
+
+  type t = {
+    io : IO.t;  (** The disk file handler *)
+    append_io : string -> unit;  (** Pre-allocated [IO.append io] closure *)
+    mutable hashset : int63 Small_list.t Array.t;
+        (** Hashset of offsets of entries in [io], keyed by the hash of the key
+            stored at each offset. Length is always a power of two. *)
```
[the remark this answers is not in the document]
CraigFe: I think of it as a `(key, value) hashtbl` with support for external storage of the key-value pairs. (I'm not sure about `(key, offset) hashtbl` as the keys are not recoverable except via the offsets.) Shall we go with that?

src/log_file.ml
```diff
@@ -0,0 +1,212 @@
...
+    Array.to_seq t.hashset
+    |> Seq.flat_map (fun bucket ->
+           let arr =
+             Small_list.to_array bucket
+             |> Array.map (fun off ->
+                    let n =
+                      IO.read t.io ~off ~len:Entry.encoded_size t.scratch_buf
+                    in
+                    assert (n = Entry.encoded_size);
+                    Entry.decode (Bytes.unsafe_to_string t.scratch_buf) 0)
```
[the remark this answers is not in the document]
CraigFe: It is, thanks. Fixed.

src/log_file.ml
```diff
@@ -0,0 +1,208 @@
...
+    let r = IO.read t.io ~off ~len:Entry.encoded_size t.scratch_buf in
+    assert (r = Entry.encoded_size);
+    Entry.decode (Bytes.unsafe_to_string t.scratch_buf) 0
+
+  let elt_index t key =
+    (* NOTE: we use the _uppermost_ bits of the key hash to index the bucket
+       array, so that the hashtbl is approximately sorted by key hash (with only
+       the entries within each bucket being relatively out of order). *)
+    let unneeded_bits = Key.hash_size - t.bucket_count_log2 in
+    (Key.hash key lsr unneeded_bits) land ((1 lsl t.bucket_count_log2) - 1)
```
Ngoguey42, line 79: If index user's keys aren't uniformly distributed they won't have correct performances with the log. Can you mention this in index's documentation?

Ngoguey42 (approved):
LGTM

src/log_file.ml
CraigFe: AFAIK this is actually already true because of the way the fan-out works. There's a warning about it here: https://github.com/mirage/index/blob/201c5a66871c41467af0a360dc0ad5d2c28ac3b1/src/data.ml#L15.

src/index.ml
```diff
@@ -912,40 +870,20 @@ struct
...
-            | None -> None
-            | Some index ->
-                let buf = Bytes.create Entry.encoded_size in
-                let n =
-                  IO.read index.io ~off:Int63.zero ~len:Entry.encoded_size buf
-                in
-                assert (n = Entry.encoded_size);
-                Some (Entry.decode (Bytes.unsafe_to_string buf) 0)))
+    | None -> true
+    | Some log -> Option.is_none t.index && Log_file.cardinal log = 0
```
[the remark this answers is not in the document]
CraigFe: Good point. I interpret this as a RO index always appearing `empty` until it does an explicit `sync` (as `find` will not be able to reach anything until then either). As it happens, we only need this for RW instances right now; would you rather add an assertion that that's the case?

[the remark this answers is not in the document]
Ngoguey42: > would you rather add an assertion that that's the case?
Yes :)

pascutto (approved):

src/index.ml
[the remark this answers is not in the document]
CraigFe: Done :+1:

PR #358, opened 2021-10-10
index: avoid unsafe buffer sharing in `Log_file`

CraigFe:
https://github.com/mirage/index/pull/355 introduced a small optimisation to re-use a "local" scratch buffer when decoding values from the log file.

This is actually unsafe: during the merge, the asynchronous merge thread and the main writer thread can attempt concurrent reads from the log file, causing contention over the scratch buffer. This can be observed by inserting `Thread.yield` calls inside the `Value.decode` implementation and then stress-testing the interface (e.g. by running the replay benchmarks).

This commit ensures that each call to a `Log_file` function gets its own scratch buffer, ensuring safe concurrent access without introducing potential contention issues.

_Note: I initially tried to just wrap `t.scratch_buf` in a mutex (https://github.com/mirage/index/pull/358/commits/0a7b05c85339943a02a9fc804976fdfa878999a5), but this was a ~5% performance regression in the replay benchmarks due to contention over the lock. The alternative of just allocating on entry to `Log_file` has a relatively small cost._

### DIFF
--- a/src/log_file.ml
+++ b/src/log_file.ml
@@ -30,6 +30,13 @@ module Make (IO : Io.S) (Key : Data.Key) (Value : Data.Value) = struct
           Entry.encoded_size)
   end
 
+  module Scratch = struct
+    type t = { buffer : bytes  (** [Bytes.length buf = Entry.encoded_size] *) }
+    [@@unboxed]
+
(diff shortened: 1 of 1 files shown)


samoht:
What's the mutex impact on the performance? Wouldn't it be better to have a scratch buffer by thread?

Issue #359, opened 2021-10-10
Add an LRU to cache `find`s


CraigFe:
Following https://github.com/mirage/index/pull/355, we have some extra memory to play with w.r.t. caching the results of `find` operations (and it's also potentially more important that we _do_ cache them, as all successful `find` operations must currently read from disk).

Here's some initial statistics on LRU cache hits achieved when running the replay benchmarks for Index:

```
$ dune exec bench/replay.exe ~/data/trace_544766125.repr -- --ops 100_000_000 
```

<div align="center"><img src="https://user-images.githubusercontent.com/26125912/136709421-be614a5d-7b5a-4457-a3b5-7ac0ac456cdf.png" width=600px /></div>
<br/>

I interpret these results as saying:

- it's relatively easy to get an LRU that performs well on this workload, needing only a capacity of ~3000 to cache nearly half the finds;
- it's not worth pushing the capacity much higher than this, as sizes of the order `log_size` have marginal gains.

We can probably pick a default capacity of the order ~30,000 elements and get good results on most workloads. This is much less than the existing log size of 500,000, so we don't need to worry about the LRU taking up too much space.

PR #366, opened 2021-10-27
index: use an LRU to cache reads

CraigFe:
Fix https://github.com/mirage/index/issues/359.

Following the recent change to the log-file implementation, successful reads of the index must always read from disk (because even log file entries are not kept entirely in memory). This traded some speed performance in favour of much lower memory usage.

For workflows with temporal locality, it's possible to recover some of this speed by adding a relatively small LRU to cache `find`s from disk. This is implemented by this commit.

**TODO**:
- [x] set the LRU size to be 0 in some of the existing tests that expect to exercise the log and data IO (right now, the tests have been weakened by having most `find`s go via the LRU).
- [x] and add some new tests that check the operation of the LRU specifically.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
 - The benchmarks now use `tezos-base58` instead of `tezos-context-hash` (#367)
 
+- Add an LRU to cache the result of `Index.find` operations. The default LRU
+  capacity is 30_000 entries. (#366)
+
 # 1.4.2 (2021-10-15)
 
 ## Fixed
--- a/index.opam
+++ b/index.opam
@@ -35,6 +35,7 @@ depends: [
   "alcotest" {with-test}
   "crowbar"  {with-test & >= "0.2"}
   "re"       {with-test}
+  "lru"      {>= "0.3.0"}
 ]
 synopsis: "A platform-agnostic multi-level index for OCaml"
 description:"""
--- a/src/dune
+++ b/src/dune
@@ -2,6 +2,6 @@
  (public_name index)
  (name index)
  (libraries logs fmt stdlib-shims mtime cmdliner logs.fmt logs.cli fmt.cli
-   fmt.tty jsonm progress repr ppx_repr optint)
+   fmt.tty jsonm progress repr ppx_repr optint lru)
  (preprocess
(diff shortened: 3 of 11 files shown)


tomjridge:
Lru stats recording fix, plus some more tests, here: https://github.com/tomjridge/index/tree/lru

CraigFe:
Thanks @tomjridge. I've pushed both changes & refactored the tests a bit but kept the logic mostly as-is (just `ocamlformat`, splitting up test cases, using more Alcotest assertions etc.)

tomjridge:
LGTM!

icristescu (approved):
Thanks!

Ngoguey42 (approved):

hannesm:
Sorry for being late to the party ;)

I have been using Lru.M earlier, and backed out of that (see https://github.com/mirage/mirage-tcpip/pull/423 https://github.com/mirage/ocaml-dns/pull/256) -- while the operations are faster (since it uses a Hashtbl), the issue is that at [`create`](https://github.com/pqwy/lru/blob/v0.3.0/src/lru.ml#L245) a Hsahtbl with the given capacity is allocated (see [Hashtbl.create](https://github.com/ocaml/ocaml/blob/4.13/stdlib/hashtbl.ml#L76)) -- which in turn allocates an Array of the given capacity.

Just in case you observe or are worried about memory consumption - since you wrap the LRU opam package, you may want to redefine your `let create cap = let lru = LRU.create 7 in LRU.resize cap lru` (i.e. initially pass a smaller number to LRU.create and then call `resize` which does not lead to any function call on Hashtable).

CraigFe:
Thanks @hannesm for the heads up. We'll keep an eye on start-up memory usage and see if this is a problem. My first thought was that it's unlikely to matter since the writer is likely to completely saturate the LRU when filling the log between merges, but this isn't true of RO instances (and it's possible that there will be many of those, potentially for relatively small stores), so this indeed may be an issue.

As it happens, we plan to move away from the `lru` package towards [`cachecache`](https://github.com/pascutto/cachecache/) relatively soon (as the implementation is a bit more efficient), but this currently has the same limitation w.r.t. initial size as `lru` so your point is still one to keep in mind.

PR #369, opened 2021-11-02
Cache fanout regions

samoht:


samoht:
Closing as the read amplification seems worse

PR #370, opened 2021-11-05
Add Documentation.md

tomjridge:
High-level doc of Index functioning

### DIFF
--- a/Documentation.md
+++ b/Documentation.md
@@ -1,0 +1,46 @@
+# Index: overview of what it is and how it works
+
+
+
+Index is essentially a key-value store. The main functionality is described in the file `src/index_intf.ml`. However this file does not describe the implementation in much detail. The purpose of this document is to fill that gap.
+
+
+
+## Initial comments
+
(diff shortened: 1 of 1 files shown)


Documentation.md
```diff
@@ -0,0 +1,46 @@
...
+
+
+
+## Initial comments
+
+Index is parameterized by various things. The most important point is that **keys and values must be fixed size**. And, indeed, in order for things to run smoothly the keys and values must be reasonably small. The Tezos instance, for example, uses 32 bytes for the keys and 3 integers (or similar) for the values. So, Index is not a generic key-value store (although it would be fairly simple to build such a thing on top of index). 
+
+Further, keys must be hashable (with the hash represented as an int), and the implementation even requires that the user specify the number of bits in the hash that are relevant.
+
+FIXME There is a worrying comment regarding the key hash: "underestimation [of the number of relevant bits] will result in undefined behavior"; most code that uses hashes should still work (albeit very slowly) even if all hashes have the same value. So, this comment is a bit unusual. What happens if hashes collide? What do we do about this?
```
Ngoguey42, line 15: You must be referring to that part of the doc:

https://github.com/mirage/index/blob/fe5e962534d192389484ecc63a25e4ab4668a2f0/src/data.ml#L17-L22

`hash_size` is used by the fan to select the bits in the short hash of a keys, depending on the size of the fan.

Overestimating imply that the fan will refer to bits that don't carry informations (typically zeroes).

Underestimating should be fine in my opinion.

pascutto: Underestimating causes an undefined behaviour because that means the bits selection drops the MSBs, which breaks the order. Having the bit selection preserving the hash order is an important invariant to the fan-out.

PR #372, opened 2021-11-08
Strengthen test of `Index.filter (fun _ -> false)`

CraigFe:
This exposes a bug whereby `Index.filter` creates a empty data file that is still loaded, breaking an internal variant that only non-empty data files should be available (and crashing an attempted `find` in the data file).

### DIFF
--- a/test/unix/main.ml
+++ b/test/unix/main.ml
@@ -924,8 +924,9 @@ module Filter = struct
   (** Test that all bindings are removed when using [filter] with a false
       predicate. *)
   let filter_all () =
-    let* Context.{ rw; _ } = Context.with_full_index () in
+    let* Context.{ rw; tbl; _ } = Context.with_full_index () in
     Index.filter rw (fun _ -> false);
+    Hashtbl.iter (fun k _ -> Index.check_not_found rw k) tbl;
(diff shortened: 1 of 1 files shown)


tomjridge:
Is the invariant "only non-empty data files should be available" documented in the code? How does it affect find? (Do we need this invariant?)

CraigFe:
> Is the invariant "only non-empty data files should be available" documented in the code?

It's mentioned [here](https://github.com/CraigFe/index/blob/filter-false-bug/src/index.ml#L116-L119), though the unclear wording is evidence that we didn't think of `filter` potentially emptying the read-write instance's data file at the time it was written. `clear` similarly sets `t.index <- None` in the read-write instance.

> How does it affect find?

The interpolation search crashes with an assertion failure when run over an empty data file, as shown in the CI on this PR.

> (Do we need this invariant?)

Potentially not if the former issue were fixed. The idea behind the invariant is to have a canonical representation of the empty data file, and since the data file is created lazily as the result of the first `merge` then this canonical representation is currently "not existing in the root directory". An advantage is that newly-created instances can use `IO.exists index_data_path` to decide if it's worth keeping an open FD for the data file.

PR #380, opened 2022-02-08
Add docs for src/unix/raw.mli

tomjridge:


src/unix/raw.ml
```diff
@@ -18,6 +18,8 @@ let really_write fd fd_offset buffer buffer_offset length =
   in
   aux fd_offset buffer_offset length
 
+(* NOTE the return value is the number of bytes read; this is not completely obvious from
+   the code *)
```
Ngoguey42, line 22: The returned value is the new offset, after the values read

src/unix/raw.mli
```diff
@@ -1,32 +1,75 @@
...
+- {b offset}: a 64-bit integer, denoting the length of the file containing valid data;
+  typically data is written to the end of a file, but not considered "valid" until a later
+  fsync completes; in this module, there is no relation between the {!fsync} function, and
+  the offset field - the offset field is uninterpreted and it is up to users of this
+  module to set the offset appropriately
+
+- {b version}: an 8-byte version string; in this module we don't impose any restrictions
+  on this field, but elsewhere it is typically used to distinguish different formats of
+  the file data (see irmin-pack's [Version] module - were the way to encode objects was
+  changed from [`V1] to [`V2]; [Version.set t s] will use exactly the first 8 bytes of the
```
Ngoguey42, line 16:
```suggestion
  the file data (see irmin-pack's [Version] module - where the way to encode objects was
  changed from [`V1] to [`V2]); [Version.set t s] will use exactly the first 8 bytes of the
```

src/unix/raw.mli
```diff
@@ -1,32 +1,75 @@
...
+  as a counter to indicate that other RO processes need to reload data
+
+- {b fan}: a 64-bit length field, followed by a string containing that many bytes; again,
+  this is uninterpreted in this module, but is used by the Index's fan; note that this
+  field means that the header does not have a fixed size
+
+How does the user know where the header ends and the file data begins? One option is to
+call {!fstat} after initializing the header info. In fact, in {!Index_unix} there is a
+separate computation of the header size: [let header = eight ++ eight ++ eight ++ eight ++
+fan_size]; presumably other users of this module do the same.
```
Ngoguey42, line 30: Yes. See https://github.com/mirage/irmin/blob/main/src/irmin-pack/IO.ml#L38

src/unix/raw.mli
```diff
@@ -1,32 +1,75 @@
...
+    data. *)
+
 val fsync : t -> unit
 val close : t -> unit
 val fstat : t -> Unix.stats
 
+(** An exception that can be raised when {b reading} (not writing!) header fields, in case
+    the amount of bytes requested for a field is not the amount of bytes actually read
+    from file. This presumably happens only for files that are not initialized correctly,
+    i.e., they do not contain a full header. *)
```
Ngoguey42, line 63: This can happen in several other situations
```suggestion
    e.g., they do not contain a full header. *)
```

src/unix/raw.mli
```diff
@@ -1,32 +1,75 @@
...
+(* FIXME rename this exception? *)
```
Ngoguey42, line 64: I think that documenting it is enough

Ngoguey42 (commented):
Thanks for documenting all this :-)

Can you pass ocamlformat so that we can make that PR mergeable?

PR #384, opened 2022-03-17
Expose Index.v_no_cache

tomjridge:
This is needed by layers for example, where in the worker we need to open a new irmin
pack+index, making sure we DO NOT reuse any existing instances.

I was never sure why there were caches baked in to the constructors of index and irmin pack stores in the first place, so please let me know if somehow this breaks some invariants I am not aware of.

The next step is to then add a similar constructor to irmin pack stores.

### DIFF
--- a/src/index.ml
+++ b/src/index.ml
@@ -622,8 +622,8 @@ struct
 
   (* Defaults shared between {!v} and {!v_no_cache}. *)
   module D = struct
-    let flush_callback,fresh,readonly,throttle,lru_size = 
-      (fun () -> ()),false,false,`Block_writes,30_000
+    let flush_callback, fresh, readonly, throttle, lru_size =
+      ((fun () -> ()), false, false, `Block_writes, 30_000)
   end
 
   let v ?(flush_callback = D.flush_callback) ?(cache = empty_cache ())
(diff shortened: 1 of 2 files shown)


Ngoguey42:
This looks good. I don't think it creates problems.

You could also have added a parameter to the existing `v` function, but adding `v_no_cache` seems fine too.

The CI is complaining because of code formatting. Could you apply ocamlformat?

CraigFe:
Perhaps an alternative approach here could be to use a `Cache.S` implementation that respects the process ID, which could then live in `irmin-pack` itself:

```diff
diff --git a/src/irmin-pack/unix/pack_index.ml b/src/irmin-pack/unix/pack_index.ml
index 055cc1c44a..d3d0b2d417 100644
--- a/src/irmin-pack/unix/pack_index.ml
+++ b/src/irmin-pack/unix/pack_index.ml
@@ -50,7 +50,30 @@ module Make (K : Irmin.Hash.S) = struct
   end

   module Stats = Index.Stats
-  module Index = Index_unix.Make (Key) (Val) (Index.Cache.Unbounded)
+
+  (** An instance cache that is additionally keyed by process ID, to ensure that
+      the index constructor is fork-safe (as is required by the asynchronous
+      worker in the layered store). *)
+  module Unbounded_instance_cache : Index.Cache.S = struct
+    module Pid : sig
+      type t
+
+      val get : unit -> t
+    end = struct
+      type t = int
+
+      let get = Unix.getpid
+    end
+
+    type ('k, 'v) t = ('k * Pid.t, 'v) Hashtbl.t
+
+    let create () = Hashtbl.create 0
+    let add t k v = Hashtbl.add t (k, Pid.get ()) v
+    let find t k = Hashtbl.find_opt t (k, Pid.get ())
+    let remove t k = Hashtbl.remove t (k, Pid.get ())
+  end
+
+  module Index = Index_unix.Make (Key) (Val) (Unbounded_instance_cache)
   include Index

   (** Implicit caching of Index instances. TODO: Require the user to pass Pack
```

Adding a `_no_cache` constructor to Index obviously works too if you'd rather do that.

tomjridge:
Hi @CraigFe. I understand your alternative, but think I prefer `_no_cache`. 

I have a question though: As I understand it, the internal caching in index and pack store (and elsewhere?), based on path names, is so that the various stores (node,commit,content) that are constructed using an underlying pack store, can all share the same pack store instance? Is this the only reason for this caching?

CraigFe:
> Is this the only reason for this caching?

It's exploited within `irmin-pack` as you point out, but Irmin (or at least its test suite) also expects `Store.Repo.v` to be memoised such that two handles build with the same config can interoperate. (This is why even purely in-memory backends have memoised constructors.)

The Tezos store also uses instance caches in a similar manner in a few places ([here](https://gitlab.com/tezos/tezos/-/blob/26ba6e034ac49270fd3e09fd72c5a44b1e85a7b1/src/lib_store/floating_block_index.ml#L104) and [here](https://gitlab.com/tezos/tezos/-/blob/master/src/lib_store/cemented_block_store.ml#L35-41)), but I'm not sure if this is actually being relied upon for correctness.

tomjridge:
@CraigFe Thinking about it more, I'm coming round to your caching-by-pid solution!

PR #386, opened 2022-04-14
Support all versions of cmdliner

icristescu:
similar to https://github.com/mirage/irmin/pull/1803
also upgrade OCamlFormat


zshipko:
Is there any reason not to merge this?

PR #389, opened 2022-06-08
Make index-bench bench executable private

### LINKED ISSUE #388: index ships a public `bench` executable
It seems that in recent versions, index started shipping a binary called bench.

If this is intentionally public, prefixing its name would help avoiding conflicts. If it's not it, making it private would be best to avoid any future problems.

This has been causing troubles in projects using opam-monorepo as index is not the only one doing so, meaning vendoring it alongside those other packages causes dune to have multiple rules to build the bench executable.

I suspect this should be a private executable, used for benchmarking in index only. There are a few tricks you can use in dune to be able to refer to a private executable as if it was public. If that may help you can see how we do this in opam-monorepo:
https://github.com/ocamllabs/opam-monorepo/blob/main/test/bin/dune#L9.

Let me know if I can help with this in any way!

NathanReb:
Fixes #388 

As discussed with @icristescu and @Ngoguey42, the package is not meant to be used by external users nor is running the benchmark so we agreed that making it private was the right move here.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,11 @@
 
 - Support all version of cmdliner (#386)
 
+## Fixed
+
+- Make `index-bench`'s `bench` executable private, avoiding executable
+  name collisions in opam-monorepo projects. (#389, @NathanReb)
+
 # 1.6.0 (2022-02-12)
 
 ## Added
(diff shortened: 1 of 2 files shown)


icristescu:
Thanks!

PR #392, opened 2023-01-09
Fix removed functions in mtime.2.0.0

patricoferris:
This adds back some of the functionality that was removed in the recent `mtime` release (it tries to use functions that were in `mtime < 2.0.0` to avoid needing to add a lower bound). The functions can be inlined instead of adding them to `Import`. Note this also requires: https://github.com/craigfe/progress/pull/31


anmonteiro:
Progress has now been released.

patricoferris:
Thanks @anmonteiro seems we'll also need `metrics-unix` to work on mtime > 2 as well. I think that is just a metadata change on opam-repository https://github.com/ocaml/opam-repository/pull/23856 (the CI will let us know...).

patricoferris:
Merged, looks like CI is happy except a lower bound

metanivek:
:raised_hands: Thanks! I tidied one small thing and added a change log entry. Once CI is green, I'll merge and do a release.

PR #394, opened 2023-07-19
Open lock files with O_CLOEXEC flag

vect0r-vicall:
This simple PR aims to open lock files with the `Unix.O_CLOEXEC` flag to avoid resource leak in case of forks.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Changed
+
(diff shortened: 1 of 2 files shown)


metanivek (approved):
Thanks! Can you add an entry in `CHANGES.md`?

vect0r-vicall:
> Thanks! Can you add an entry in `CHANGES.md`?

Done!

PR #397, opened 2024-02-22
(draft) Eio backend

art-w:
Add `index.eio` to use Eio for I/O as an alternative to `index.unix` blocking I/O.

(draft because it is missing the ability to lock files, and it introduces a dumb breaking change to `index.unix` interface)


zshipko:
@art-w do you know what is needed to move this out of draft state?

Issue #398, opened 2024-05-24
Updated values are not seen in RO instances


edwintorok:
Using 'index.1.6.2' on OCaml 4.14.1:
```
$ cat >testme.ml <<EOF
#require "index.unix";;
module L = struct let length = 4 end;;
module I = Index_unix.Make(Index.Key.String_fixed(L))(Index.Value.String_fixed(L))(Index.Cache.Noop);;
let path = "testme.ocamlindex"
let t = I.v path ~log_size:16384;;
let ro = I.v path ~readonly:true ~log_size:16384;;
I.replace t "1234" "aaaa";
I.flush t; I.sync ro;;
I.find ro "1234" |> print_endline;;

I.replace t "1234" "aaab";
I.flush t; I.sync ro;;
I.find ro "1234" |> print_endline
EOF

$ utop testme.ml
aaaa
aaaa
```

Maybe I misunderstand how `Index` should be used, but it does notice that new keys are added, but then it doesn't notice that a key's value has been changed.

How can I get the updated value from `t` to `ro`, am I missing some setting/call, or is this a bug?

art-w:
It's a bug indeed, many thanks for the short repro!

PR #399, opened 2024-05-24
Fix RO sync: clear LRU for all log updates

art-w:
As reported by @edwintorok in #398, a read-only instance would sometimes fail to find the latest value after doing a `sync`, because it would keep finding the old value in its LRU cache (and ignore the more recent writes).

The quick-fix is to always clear the LRU whenever `sync` discovers that the log has changed on disk... It's not super satisfying as we could be more fine grained and update the LRU with the new bindings from the log, but it's the strategy the code was already using for some (but not all) log updates.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - Lock files are now opened with O_CLOEXEC flag (#394, @vect0r-vicall)
 - Update to cmdliner.1.1.0 (#382, @MisterDA)
 - Mirage support: optional dependency to unix (#396, @art-w)
+- Fix RO sync: clear LRU for all log updates (#399, @art-w)
 
 # 1.6.2 (2023-06-06)
 
--- a/bench/dune
+++ b/bench/dune
@@ -26,7 +26,7 @@
  (preprocess
   (pps ppx_repr))
  (libraries index index.unix unix cmdliner logs repr ppx_repr common
-   tezos-base58 optint fmt rusage mtime mtime.clock.os digestif))
+   digestif.ocaml tezos-base58 optint fmt rusage mtime mtime.clock.os))
 
 ;; Require the above executables to compile during tests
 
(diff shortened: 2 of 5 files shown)


edwintorok:
Clearing all entries is probably safer indeed. Another approach might be to have a generation counter on each entry, and then compare the generation counter of the file with the generation counter of the cached entry, and consider the entry stale if it is older.

art-w:
Hmm I'm slightly worried of the performance impact of double-checking the LRU freshness on every query, it felt more predictable to only penalize the read-only instance for calling `sync` by clearing its LRU. Wouldn't the generation counter be equivalent to an LRU clear as it would invalidate all entries that are older than the latest update? (note that we only call `Lru.clear` if `sync` sees that an update happened, but otherwise the LRU is preserved)

That being said, I believe it should be possible for `sync` to update the LRU with the newer values read from the log, but the code change is more involved... Perhaps as a follow-up PR, unless you are already seeing poor performances caused by the LRU clear?

edwintorok:
> Hmm I'm slightly worried of the performance impact of double-checking the LRU freshness on every query, it felt more predictable to only penalize the read-only instance for calling `sync` by clearing its LRU. Wouldn't the generation counter be equivalent to an LRU clear as it would invalidate all entries that are older than the latest update? (note that we only call `Lru.clear` if `sync` sees that an update happened, but otherwise the LRU is preserved)

You are right, keeping a RO client efficiently in-sync with another one is not really possible currently, and the generation counter wouldn't really help (you need to discard all LRU entries, whenever *anything* changes in the index, even if we have per-entry generation counters, we'd still need to read that counter to know whether it changed or not).

> 
> That being said, I believe it should be possible for `sync` to update the LRU with the newer values read from the log, but the code change is more involved... Perhaps as a follow-up PR, unless you are already seeing poor performances caused by the LRU clear?

Yes, that sounds like a better approach than what I was suggesting with the generation counters. Followup PR sounds good, currently I'm just evaluating whether I can use 'index' in a project, I'm not using it yet as such.

PR #400, opened 2025-03-11
tests/unix: add `logs.threaded` as a dependency

anmonteiro:
- The upcoming release of logs.0.8.0 (https://github.com/ocaml/opam-repository/pull/27602) installs `logs.threaded` in `<dst-dir>/logs/threaded`
- Previously the tests in `tests/unix` would only find this library by chance, since `logs_threaded.cm*` was installed in the same directory as the `logs` library

### DIFF
--- a/test/unix/dune
+++ b/test/unix/dune
@@ -1,5 +1,5 @@
 (tests
  (names main force_merge io_array)
  (package index)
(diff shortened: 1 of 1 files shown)


art-w:
Many thanks! (the CI will be fixed by #402 as other packages have had similar breaking changes)

